PR #50, opened 2013-03-26
Change 'Viet Nam' to 'Vietnam'.

allika:
Why 'Viet Nam' in not Vietnam?


frenkel:
It seems https://www.quora.com/Is-it-Vietnam-or-Viet-Nam suggests Vietnam as well. What is the source of the current 'Viet Nam' spelling?

PR #98, opened 2013-08-27
Add GEC (FIPS-10) code

lorddoig:
GEC/FIPS-10 has been deprecated, but I needed it so the world might as well have it.

Sorry about the formatting of the data file, there was no way I was adding these codes by hand and some of the formatting got changed.  I wrote a nasty little spec to recursively check that I hadn't changed any country's attributes (not here) and it's all fine.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -61,6 +61,7 @@ Country Info
     c.number #=> "840"
     c.alpha2 #=> "US"
     c.alpha3 #=> "USA"
+    c.gec    #=> "US"
 
   Names & Translations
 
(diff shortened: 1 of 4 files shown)


rposborne:
@hexorx Any input on this?

hexorx:
I have no problem with extra information like this being added.

rposborne:
@lorddoig  could we get this updated to latest?

lorddoig:
To be honest I'm not sure where I got this info from or how I went about adding it. Updates to GEC coding are published [here](http://earth-info.nga.mil/gns/html/gazetteers2.html) but it's not (easily) machine readable and actions required here resultant from the changes they make aren't immediately obvious.

Unless there's proven demand from others I'd be tempted to drop this owing to added complexity and maintenance requirements (~2 updates/year).  Thoughts?

rposborne:
@lorddoig yup I took care of it.   Thanks for the contrubution

PR #144, opened 2014-05-01
Fix Barbados english name

zohrehj:
Barbados/Barbade english name should be "Barbados"


rposborne:
closed in favor of #153

PR #145, opened 2014-05-08
adding 11 to national_number_lengths on Brazil because Sao Paulo city us...

ustrajunior:
In São Paulo (Brazil) the telephone number has one more digit than in other cities.


rposborne:
yup :+1:

PR #148, opened 2014-05-28
Removed Republic of China reference from Taiwan

johnrees:
I'm getting tired of offended Taiwanese people giving me grief over this one!


rposborne:
Isn't that the country's official name? http://en.wikipedia.org/wiki/Taiwan

PR #158, opened 2014-06-26
Fix "warning: already initialized constant COUNTRIES"

Davidslv:
This should not harm the gem in anyway, its actually to prevent the following warning:

```shell
countries-0.9.3/lib/iso3166.rb:7: warning: already initialized constant COUNTRIES
```


rposborne:
This shouldn't be necessary.  If this issue gets more traction we will consider it.

PR #168, opened 2014-08-07
Add country names used by Stripe as alternate names

bprotas:
We use the countries gem to look up information for addresses we get from Stripe, which sometimes has funny punctuation (or lack of) or misspellings in the country names; this PR adds all the country names that Stripe uses as alternate names in the appropriate entries.


rposborne:
Could you please rebase this onto master?

PR #177, opened 2014-09-10
Fix Country#equality method

estum:


rposborne:
This is disgustingly elegant.

PR #178, opened 2014-09-12
Fix mongoid field behavior

gauthier-delacroix:
Current mongoid Country field type behavior makes invalid country values raise a `BadMongoidTypeError` exception on assignation.
This is not the behavior of other Mongoid types ; actually `BadMongoidTypeError` error only exists in this country gem...
When you assign invalid values on base Mongoid types, like `Integer`, no exception is raised. The value is just set to the closest approximation: an integer field is just set to zero when assigning a string.

This PR brings this standard behavior to the Country field: when an invalid value is assigned, the field is set to `nil`, with no exception.

If you want to ensure the validity of the field content, just validate the field presence: if it's not empty, then it's a valid country !

### DIFF
--- a/lib/countries/mongoid.rb
+++ b/lib/countries/mongoid.rb
@@ -1,7 +1,6 @@
 module ISO3166; end
 
 class ISO3166::Country
-  class BadMongoidTypeError < StandardError; end
 
   def mongoize
     ISO3166::Country.mongoize(self)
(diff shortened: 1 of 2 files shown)


rposborne:
It's quite common to see "Upgrading to 1.x" in gem readme's, adding a note in that regard if this breaks any API is certainly prudent. I prefer not to muddy the output using ```post_install_message```

PR #182, opened 2014-09-30
adding Nagorno Karabakh information

lumir:


rposborne:
Nagorno Karabakh doesn't appear to be a UN recongized countries @hexorx do we have a stance on this type of data?

hexorx:
While we add extra information about countries in ISO 3166 I don't think we should add information for countries that are not part of the standard.

PR #191, opened 2014-11-26
add decimal coordinate for each country

jabawack81:
I've added the decimal latitude and longitude of each country so it can be used as data source for http://jvectormap.com/

### DIFF
--- a/lib/countries/country.rb
+++ b/lib/countries/country.rb
@@ -29,7 +29,9 @@ class ISO3166::Country
     :languages,
     :nationality,
     :eu_member,
-    :postal_code
+    :postal_code,
+    :latitudeDec,
+    :longitudeDec
   ]
 
   AttrReaders.each do |meth|
--- a/lib/countries/country.rb
+++ b/lib/countries/country.rb
@@ -30,8 +30,8 @@ class ISO3166::Country
     :nationality,
     :eu_member,
     :postal_code,
-    :latitudeDec,
-    :longitudeDec
+    :latitude_dec,
+    :longitude_dec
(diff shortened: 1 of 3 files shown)


rposborne:
Looks like you used camelcase for this,  could you please use Ruby style snake case.

jabawack81:
My bad, renamed.

rposborne:
@jabawack81 we just did a reasonable amount of restructuring data,  any chance you could move this over to the new data format? Each country has a ```lib/data/countries/US.yaml``` one file per country

rposborne:
Here is a rake task I used on a few PR's 
``` ruby
task :pull_new_keys_from_old_file do
  require 'yaml'
  countries = YAML.load_file File.join(File.dirname(__FILE__),  'countries.yaml')
  country_list = []
  countries.each do |cc, _data|
    keys = [ 'min_longitude',
            'min_latitude',
            'max_longitude',
            'max_latitude'
           ]
    country_data = YAML.load_file File.join(File.dirname(__FILE__), 'lib', 'data','countries', "#{cc}.yaml")
    keys.each do |key|
      country_data[cc][key] = countries[cc][key]
    end
    File.open(File.join(File.dirname(__FILE__), 'lib', 'data', 'countries', "#{cc}.yaml"), 'w') do |h|
      h.write(country_data.to_yaml)
    end
  end
end
```
Just put the keys you want to pull in the keys array, and have your old countries file at the root of the rep

jabawack81:
thi PR can be deleted I've made a new one with the new structure.

Jaba

PR #199, opened 2015-01-02
Add Russian as Estonian language

martinholmin:
Russian is very common in Estonia.


rposborne:
29.7% of people in Estonia speak Russian.  I think this is a practical addition.

PR #204, opened 2015-01-19
Fix spelling

styx:
`Белоруссия` is wrong country name. The right official name is 'Респулика Беларусь' or in short 'Беларусь'.


rposborne:
Closed in favor of #184

styx:
@rposborne, great, thnx.

PR #211, opened 2015-01-23
i18n_data  + Countries into individual files

rposborne:
Splitting countries into smaller individual files (all loaded on class load)
Use i18n_data for country names and translations.
Removed Translations key from existing files.

For @hexorx's humble consideration.


lib/countries/country.rb
```diff
@@ -1,9 +1,10 @@
 module ISO3166; end
 
 class ISO3166::Country
-  Data = YAML.load_file(File.join(File.dirname(__FILE__), '..', 'data', 'countries.yaml')) || {}
-  Names = Data.map {|k,v| [v['name'],k]}.sort_by { |d| d[0] }
-  NameIndex = Hash[*Names.flatten]
+  Codes = YAML.load_file(File.join(File.dirname(__FILE__), '..', 'data', 'countries.yaml')) || {}
+  Data = {}
+  Codes.map { |alpha2| Data[alpha2] = YAML.load_file(File.join(File.dirname(__FILE__), '..', 'data', 'countries', "#{alpha2}.yaml"))[alpha2] || {} }
```
hexorx: This should probably be Codes.each instead of map.

hexorx:
Great work, this will make managing the data much easier.

lib/countries/country.rb
rposborne: ... Yup, but doesn't everybody love orphaned objects?

PR #220, opened 2015-01-28
Multibyte chars support

kenjione:
Fix for searching by lowercase multibyte characters (non-latin languages)

```ruby
>> Country.find_country_by_name("Россия")
#<Country:0x007fcbc826bec0 @data={...} >
>> Country.find_country_by_name("россия")
nil


rposborne:
@hexorx, how do you feel about adding activesupport as a dependency.

hexorx:
I would like to avoid it if possible. The common use case for the gem is in rails but it can be used outside of rails and active support is pretty heavy.

rposborne:
Closed in favor of #337

PR #221, opened 2015-01-28
Fix address format for Spain

stayhero:
According to this document the region for address labels in Spain has to
be printed in a separate line:

http://www.upu.int/fileadmin/documentsFiles/activities/addressingUnit/espEn.pdf

### DIFF
--- a/lib/data/countries/ES.yaml
+++ b/lib/data/countries/ES.yaml
@@ -4,7 +4,8 @@ ES:
   address_format: |-
     {{recipient}}
     {{street}}
-    {{postalcode}} {{city}} {{region}}
+    {{postalcode}} {{city}}
+    {{region}}
     {{country}}
   alpha2: ES
   alpha3: ESP


stayhero:
Furthermore I was wondering if there should be another template variable for the region code. Because in cases where the region should be printed on a separate line, the full region name should be used. Like:

> Jone Doe
> Plaza de las Descalzas 27
> 29400 MARBELLA
> MÁLAGA
> SPAIN


Whereas in most cases where the region should be printed on the same line, the abbrevation should be used, like

> MR JOHN DOE
> 123 MAGNOLIA ST 
> HEMPSTEAD NY 11550–1234 
> UNITED STATES OF AMERICA

Having only the placeholder {{region}} would mean during parsing of the template one has to deceide to either use the full state name (New York) or the abbrevation (NY). 

What do you think?

rposborne:
I think that is a prudent addition, something along the lines of "region_short"?  Though we should have a firm version bump for this.

stayhero:
@rposborne Yes, I guess that would be necessary. 

Of course it would be possible to change the contents of the address formats for countries which require the short version to use {{region_short}} or similar. And country.address_format could return some kind of liquid code to automatically set this value from {{region}}. Like:

```{% unless region_short %}{% assign region_short=region %}{%endunless%}```

But I guess not everyone is parsing the adress_format with Liquid but rather simply replace the placeholders? And of course the output of address_format would not look as nice as now.

**Comment by rposborne** [comment]
Yea I don't think we want to switch to liquid.   I think this gem focus should really be just static data, anything that moves us towards the facts is a good thing,  I think we should just have a region_short and that be that.   We will bump the gem hard to a major revision number.  Would you mind adding a PR for this, and a clarification in the readme?

**Comment by stayhero** [comment]
Another option would be to create a new method for Country to let this parsing happen automatically. Something like this:

```ruby
address = Country.new('US').formatted_address(recipient: 'John Doe', street: '123 Main Str.', ...)
```

stayhero:
Sure. If you say this would be merged into master I would work on a pull request, changing all address_format's contents to use {{region_short}} where applicable. 

A simple way to do this would be to use {{region_short}} where {{region}} is currently not on a separate line. If you wish I could verify that the correct address formats are used (e.g. in master currently for Spain it's wrong). The latter would require a couple of days for me. But I guess it's not that urgent.

rposborne:
Yea, no rush whatsoever.  I would do the change programmatically to start, and let PR's come in for corrections. I am still searching for a good data source for address formating.

stayhero:
At https://www.deutschepost.de/de/b/briefe-ins-ausland/brief-beschriften.html the german postal service links to documents from the "Universal Postal Union":

http://www.upu.int/en.html
e.g.
http://www.upu.int/fileadmin/documentsFiles/activities/addressingUnit/fraEn.pdf

I assume it's a reputable source but I'm not sure about the copyright. At least they use a small &copy; symbol on each document. :crying_cat_face:  Not sure if that kind of information is copyrightable. Probably it's wiser to use another source or build our own database my manually looking up for the desired information with the local postal website.

PR #222, opened 2015-01-28
Add provinces as an alias for subdivisions

cllns:
There already exists an alias for `.subdivisions` as `.states` 

I think there should be `.provinces` as well!

I added a test for this, and also fixed the existing test for `.states` (I think... It used to test against itself)

### DIFF
--- a/lib/countries/country.rb
+++ b/lib/countries/country.rb
@@ -81,6 +81,7 @@ class ISO3166::Country
   end
 
   alias :states :subdivisions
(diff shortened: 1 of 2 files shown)


rposborne:
I would actually like to remove the states alias, as it's simply clutters the API.  Everything is a subdivision.  That should be all we need.

cllns:
I like the aliases but I understand your reasoning!

Issue #223, opened 2015-01-30
Performance regression with country_select


stayhero:
Hi. I wanted to use the country_select gem with the latest countries version, and there is a performance regression because of the translated country names. See https://github.com/stefanpenner/country_select/issues/92

Obviously it would help the country_select gem to have a country.translated_name(locale) method. Currently it uses 

```ruby
country.translations[I18n.locale.to_s]
```

to get the translated country name. As the country_select gem needs to fetch the name for each of the 249 countries, and countries gem loads ~92 translations, it results in 23250 calls to ISO3166::Country#translations and takes 700ms (on a dev machine) to load the list.

Would you accept a pull request for implementing a method like country.translated_name(locale)? I see also use cases for such a method at other places. Most often you don't want to retrieve all translations for a specific country.

rposborne:
I think we have one thing that might help or could be tweaked to produce better data for this purpose.   https://github.com/hexorx/countries/blob/master/lib/countries/country.rb#L124

Regardless anything related to performance we will give the time of day.

rposborne:
So a quick peek under the hood says, don't depend on this gem directly, we have extracted the translations aspect of countries upstream to [i18n_data](https://github.com/grosser/i18n_data) which pulls it's data down for the debian sources, and was much more complete then countries translations were. 

That being said, I think we should do a breaking change for the all_translated to not simple return values but the hash key ``` { 'DE' => 'Deutschland' ...} as this provides more use of the data, vs assuming how it will be used.

stayhero:
To implement the newly added translation method it would be easier to let it return nil if a translation is not found.

See pull request
https://github.com/hexorx/countries/pull/227

That would reduce the changes necessary in country_select and make it more consistent with the translations method?

stayhero:
Thanks for merging. I think this issue can be closed although to let country_select release a new version it will need to bump the dependency on countries gem to the next release of it (like 0.10.1)?

rposborne:
I think we have some marge API changes on master, so we will need to release a or at least an 0.11.  

Let me put up the bat signal.  :satellite: @hexorx

PR #224, opened 2015-01-31
Adding translation method for more performant access to specific locales

rposborne:
See #223 , and https://github.com/stefanpenner/country_select/issues/92


rposborne:
@stayhero would you look over this and see if you agree?

stayhero:
Great. :+1: 

When country_select gem uses that new method the performance is back to normal (12ms vs. 700ms). Thanks!

PR #226, opened 2015-01-31
Added Slovenian names for all countries

jernejstrasner:


jernejstrasner:
Hmm I can't seem to view build info on Shippable so I can't fix the error breaking the build. The specs run fine on my local machine though.

rposborne:
Thanks for this though I want to get a bit of clarification of the names field before I merge. See #228 since we already have all translations, I consider this names field to be the "Slang" list, or other things a country might be called that is not an any official spec.

jernejstrasner:
Yeah, I wasn't sure about it either, but I saw chinese or japanese names in there so I though I might stick that in. What I'm not clear about is how to put the names properly into the `translation` field and search based on that.

rposborne:
@jernejstrasner We currently don't have the facilities in place to augment the data coming from the i1l8n_data gem.  But I can certainly look at extending find_by_name to call upon translations.

jernejstrasner:
@rposborne That'd be nice. Because what I'm basically doing is trying to get ISO identifiers from Slovenian names for languages. I'll look into i18n_data a little bit more.

rposborne:
check out #229

PR #227, opened 2015-01-31
Return nil if translation is not found

stayhero:
The newly added translation method in ISO3166::Country should return nil
rather than throwing an I18nData::NoTranslationAvailable exception if a 
translation is not found.


stayhero:
See https://github.com/hexorx/countries/issues/223

rposborne:
Solid addition thanks for this.

PR #232, opened 2015-02-09
Taiwan is not part of China

tonytonyjan:


rposborne:
@tonytonyjan We try to conform to the ISO specs for subdivisions which does indeed list Taiwan as a part of CN.  

See (here)[http://en.wikipedia.org/wiki/ISO_3166-2:CN]

tonytonyjan:
Sorry, I didn't notice that. It's sad that I can't send pull request to ISO.

rposborne:
Yea, seriously.

PR #233, opened 2015-02-11
Expose Country only when it's safe

ixti:


rposborne:
I think I would prefer to raise an error here since we already have facilities to prevent this name conflict, then have it fail silently.

ixti:
Either way - I'm OK. Just wanted to avoid `:require => ...` in Gemfile ;))

kennethkalmer:
I'm bumping into this issue where I already have a `Country` global, and I thought to open a similar PR. However, I thought instead of magically defining a `Country` class, why not explicitly `require "countries/global"` if you need the old behavior?

rposborne:
@kennethkalmer I like the idea of just switching the default behavior vs dynamically detecting if we are going to conflict.

rposborne:
I have a PR using @kennethkalmer suggestion, here #285

ixti:
Closed in favor of #285

PR #235, opened 2015-02-17
IR.yml updated

yottanami:
The persian translation of Iran added.


rposborne:
@yottanami this looks to be the official translation,   the names section is for "slang" or official names that might be used.  e.g.  Great Britain -> United Kingdom.

PR #239, opened 2015-03-03
Added newly formed state(Telangana), after bifurcation of AP in India

tsatyadevreddy:
A new state formed in India after bifurcation of AP state. So added it.


rposborne:
Fantastic, Here is a source for future reference. http://en.wikipedia.org/wiki/ISO_3166-2:IN#cite_note-7

PR #242, opened 2015-03-11
Enabling Ruby warnings

franckverrot:
These commits are removing some of the warnings that appears when running a dependent gem/app in verbose mode (`ruby -w`).

Some of those warnings are code-smells, so that's why I'm trying to get rid of them.

Please tell me if you wanna squash those. I didn't as they're basically isolated and independent from one another.

Thanks for this gem!


rposborne:
Thanks for this @franckverrot love killing stuff like this!

rposborne:
0.11.2 released!

Issue #186, opened 2014-10-22
Subdivision tiers


madshargreave:
Hey

This page http://en.wikipedia.org/wiki/ISO_3166-2:GB specifies subdivision names as listed in the ISO 3166-2 standard

There are 4 codes ENG, NIR, SCT and WL (England, Northern Ireland, Scotland and Wales) that are higher tiered than what the gem currently returns as UK subdivisions.

This means that we have different tiers of subdivisions, where the gem currently only supports the lowest level

Will you be able to implement this, and if not, do you prefer an approach that we could take and submit a pull request?

Much appreciated
- Mads

rposborne:
I think something along these lines would be fantastic, but need some thought about the structure.   Similar to the way Google uses the "Administrative_level 1-8" in their addresses.

rposborne:
Here is a nice break down from Open Street Map, http://wiki.openstreetmap.org/wiki/Tag:boundary%3Dadministrative

rposborne:
Here are my thoughts for overall structure. 
``` yaml
administrative_level_1:
  name: region
  divisions: 
    CE:
      name: Central
      names: Central
      translations:
        en: Central
    GH:
      name: Ghanzi
      names:
      - Ghansi
      - Khanzi
```

This would be accompanied by a ```sub_division_catagories.yml``` file which would hold options of sub_divison names and their respective translations.

noyg:
Will this feature be developed?

rposborne:
Some more resources from wikipedia https://en.wikipedia.org/wiki/List_of_administrative_divisions_by_country, 

@noyg This is up on the list, but it's a pretty big lift, I am more than happy to bring other people in to help get it done if you have the time.

rposborne:
So putting a bit more of thought into this, I think for simplicities sake we can drop the `catagories.yml` file 

``` yaml
administrative_level_1:
  translations:
    en: 'Region'
    cz: 'kraj'
  divisions: 
    CE:
      translations:
        en: Central
    GH:
      translations:
        en: Central
```

I also want to strip out `name` and `names`, as its very redundant to the translations.

manhhung741:
How about we add a `type` property to `subdivisions`? I think it'll be a simple change that immediately make it more useful. For the context, I'm trying to get the list of 51 states of the US. Right now `Country[:US].states` returns all kind of subdivisions, and there is no way to filter unneeded subdivisions.

coelhone:
+1

pmor:
[...]

PR #253, opened 2015-04-10
Added coordinates for subdivisions

fabn:
I need coordinates and bounding boxes for an app I'm working on. Here's a simple proof of concept for Italy. Are you interested in this kind of pull request?

If you are I'll add missing data for all other subdivisions.

Let me know.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
 gem 'rake'
+gem 'geocoder'
+gem 'retryable'
 
 # Specify your gem's dependencies in countries.gemspec
 gemspec
--- a/Rakefile
+++ b/Rakefile
@@ -56,5 +56,48 @@ task :cache_translations do
 
   end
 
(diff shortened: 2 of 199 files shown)


rposborne:
@fabn I am interested in this sort of data.  My only thoughts would be trying to think about #186 in context.

fabn:
#186 is a very different (and hard) task... I don't know if it's possible to have a so nice hierarchy for all countries.

In any case regarding my pull request I can include the task I've used to generate those data, in this way it can be adapted and enhanced later when #186 is done.

The only downside is that it has an external dependency on [geocoder gem](https://github.com/alexreisner/geocoder), but it can be added as development dependency in gemspec and this should not be an issue.

rposborne:
Fair enough.    I am fine with any additional dev dependencies.

fabn:
Here are all data I was able to get through google geocoding api. I've also added a task in `Rakefile` used to produce these info. Final result is pretty good, only 59 states are still missing.

Let me know if this is ready to be merged, If you want I can squash it in a single commit.

rposborne:
@fabn If you squash I'll merge this in.

fabn:
@rposborne rebased and squashed.

PR #265, opened 2015-05-28
Update NA subdivisions to latest ISO3166

danshultz:
I've updated the latest Namibia subdivisions based upon http://en.wikipedia.org/wiki/ISO_3166-2:NA

There are some others that I can start updating to make sure this list is the latest and greatest but I wanted to get your feedback on this change before doing so.

Thanks!


rposborne:
:ship:

PR #267, opened 2015-06-10
Make ::all_translated return the same format as ::all, but with translated names

Folkman:
I thought it might be more intuitive to have `Country.all_translated` return with the same array list as `Country.all` since the method names are indicating the same type of response.  This makes it really useful to use for country select dropdown boxes so the same country code is saved into the database regardless of language chosen.

### DIFF
--- a/lib/countries/country.rb
+++ b/lib/countries/country.rb
@@ -119,7 +119,10 @@ class ISO3166::Country
     alias_method :countries, :all
 
     def all_translated(locale = 'en')
-      translations(locale).values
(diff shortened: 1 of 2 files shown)


rposborne:
So we have changed ::all's default behavior to no longer assume what people want in favor of just passing back ISO3166::Country objects in an array. 

We did add a method ::all_names_with_codes which indeed prefers the translated name over the name in the yaml. 

I will close for now and reopen if this does not meet your needs

Folkman:
Thanks!

PR #269, opened 2015-06-12
Murica is not France

seuros:
I'm pretty sure they are located in different continents. :)


rposborne:
hey @seuros we have a known issue of this cached data having some bad data.  We are looking at trying to find an improved data source.  :/

rposborne:
This was fixed via 6547d43c34568598587da9699c3bf81639fc7c46

PR #274, opened 2015-06-25
Fix New York, Washington, Oklahoma state's latitude/longitude information

GUI:
When Google's geocoder is passed "New York, United States", it actually returns the geocoding information for New York City, and not the actual state. So the previously stored details were for the extent of the city, rather than the state. This fixes the initial geocoding logic by making a special case for New York to geocode "New York State". The stored YAML details have been updated with the new results for the state.

I've seen Google's geocoder change how it geocodes the string "New York" over the years, but appending "State" to the end has been a reliable way to force it to geocode the state (at least in my experience for the past couple years). So I think this approach should be safe in case anyone want to do a full re-geocoding sometime in the future.

When re-running the geocoding, it looks like a couple new locations in Namibia now have location results, so I committed those results as well, but if you'd prefer I separate those out from this pull request, just let me know.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -93,7 +93,11 @@ task :fetch_subdivisions do
     # Iterate subdivisions
     state_data = c.subdivisions.dup
     state_data.reject { |_, data| data['latitude'] }.each do |code, data|
-      if (result = geocode("#{data['name']}, #{c.name}"))
+      location = "#{data['name']}, #{c.name}"
+      if(c.alpha2 == "US" && code == "NY")
+        location = "New York State, United States"
+      end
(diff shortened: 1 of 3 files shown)


GUI:
Oh, I found a couple more: Washington state's location information was for Washington DC, and Oklahoma was for Oklahoma city. Sorry for the commit churn (let me know if you'd like me to squash all this into a single commit). But I've now done a more through search of all the states in the US, and I think this covers all of them. I also experimented with geocoding everything in the US with the "State" suffix, but that causes some issues with the US territories, so I kept the code with just handling these few special cases.

PR #285, opened 2015-08-20
Memory refactor

rposborne:
Derailed Benchmarks are awesome, countries previous memory usage, less than awesome. 
 
Address object creation from yaml parser with switch to marshal parser.


dgilperez:
Wow, a great improvement already. Results with `CUT_OFF=0.5 derailed bundle:mem`:

Before:

    countries: 28.5117 MiB
      iso3166: 28.4492 MiB (Also required by: country_select)
        iso4217: 0.8711 MiB
          currencies/currency: 0.7383 MiB

After:

    countries: 1.1445 MiB
      iso3166: 1.1055 MiB (Also required by: country_select)
        iso4217: 0.9023 MiB
          currencies/currency: 0.7891 MiB

rposborne:
So I will come clean in some regards, this mainly address the memory bloat in two ways, parsing the YAML was causing a massive amount of strings, so we switched to marshaling. And we no longer do anything on boot, aka no constants etc.  It WILL consume around 8mb in a production environment after a few calls come through,   I think I will re attempt to split out a few things, so we can selectively load locales.  Cause lets be real, most use cases don't need every language.

dgilperez:
:clap: :beers:

PR #286, opened 2015-08-25
Add alternate spelling of Macao/Macau

nityaoberoi:
Macau is used by Stripe

https://en.wikipedia.org/wiki/Macau

### DIFF
--- a/lib/data/countries/MO.yaml
+++ b/lib/data/countries/MO.yaml
@@ -13,6 +13,7 @@ MO:
   name: Macao
   names:
   - Macao
+  - Macau
   - "マカオ"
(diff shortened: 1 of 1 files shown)


rposborne:
Awesome, though the names key is basically the junkdrawer of names, not official translations. 
If this should be an official translation that we are getting wrong you can add a correction here https://github.com/hexorx/countries/blob/master/lib/data/translation_corrections.yaml

nityaoberoi:
It seems like Macau can also be called Macao based on what I'm reading about it, so I don't know if we should consider it as an official translation. I.e., some services may use the Macao spelling. In that case, should that still be added to the translations?

Issue #257, opened 2015-04-21
sorting by country name


amolpujari:
`Country.order_by_country_name` 

And this should get all the countries sorted by country name.

By default it does sorting by apha2, and when used in dropdown, Germany appears after 'C' and not among 'G's

lebrauncom:
@amolpujari call `.sort` on the returned array. This will order the selected countries by country name instead of country code.
```
Country.all.sort
```

pjg:
I also recommend using https://github.com/grosser/sort_alphabetical

guilleart:
`Country.all.sort` fails in 1.0.0
```
1.9.3-p484 :003 > Country.all.sort
ArgumentError: comparison of ISO3166::Country with ISO3166::Country failed
```

rposborne:
1.1.0 has been released with @guilleart 's compare operation

PR #287, opened 2015-08-26
Adding compare operation to country

guilleart:
#257


rposborne:
Awesome could you add a test for the expected functionality? and we will get this merged in.

guilleart:
@rposborne spec done. Let me know what you think

PR #288, opened 2015-08-27
.to_s should return the translation of the current locale and fall back to name if no translation present.

rposborne:
This should let sorting occur easily on a per language basis, as name is relative.

### DIFF
--- a/lib/countries/country.rb
+++ b/lib/countries/country.rb
@@ -92,7 +92,7 @@ class ISO3166::Country
   end
 
   def to_s
-    @data['name']
+    translation(I18n.locale.to_s) || @data['name']
   end
 
   def translation(locale = 'en')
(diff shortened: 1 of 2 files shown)


rposborne:
@guilleart how do you feel about this in context of #287

guilleart:
Hmm this would work as long as the names didn't have weird characters like å é ü etc. In case there are, I don't think Ruby sorts them nicely.

I looked up some solutions on how to sort international characters and they seem to rely on gems like https://github.com/grosser/sort_alphabetical

I can see this being annoying for users using those locales, but on the other hand adding a dependency like that just for this case seems like overkill

rposborne:
You are absolutely right, I guess it boils down to where the responsibility is to prevent those sort of gotchas.

PR #289, opened 2015-08-27
Selective Translation Loading (REQUEST FOR COMMENTS)

rposborne:
This PR, allows for selection of which translations are loaded for each country.  

Primary Use case is for a site that supports English, French, German, and Spanish.  The country objects will not be built with any other translations or translated names.  

This provides major gains in  speed and memory usage,  speed as we reduce the size of objects certain queries iterate over, and memory by not loading as much.  But long story short we are at about 0.16 mb retained for 3 locales, and 2.177967 mb retained for all 92 locales. 

State is maintained via the configuration object, tracking which locales are loaded and to determine when a flush / reload needs to occur.  

One breaking change is since the creation of the complex objects is happening at run time I am no longer bothering calling uniq on the translated_names. as it appears to be creating very large amount of strings about 33mb (allocated) worth of them, so these names can no longer be assumed to be uniq.


rposborne:
Closing in favor of #317

PR #309, opened 2015-11-10
fix cache updater

heavyk:
the dir structure seems to have changed.

also, loading countries instead of i18n_data produced some strange warnings.


rposborne:
@heavyk thanks for this,   yes folder structure did change, so we didn't clobber the global load path.

PR #311, opened 2015-11-13
Adding TimeZone Support

rposborne:
This is a first draft on optional timezone support.   I could be convinced either way if this should be included by default.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 gem 'rake'
 gem 'geocoder'
 gem 'retryable'
+gem 'tzinfo'
 
 # Specify your gem's dependencies in countries.gemspec
 gemspec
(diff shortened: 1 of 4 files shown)


dgilperez:
:clap:

esaborit4code:
As this gem doesn't have any optional feature or config yet, I would add timezones by default as long as it doesn't add any performance issues or some kind of overriding conflict.

Anyway, I don't know the codebase deeply, so consider this just as an opinion.

rposborne:
@esaborit4code We are going to be adding a config come 2.0. Which we will wrap this into.  I want optional loading of all data, so we can keep our memory usage down so we can be as useful as possible across the board.

rposborne:
Closing in favor of #317,   Timezones will be extended by default, but tzinfo is not a dependency.

PR #322, opened 2016-01-07
Added XAF as currency code for Congo

espen:
Sources:
http://www.oanda.com/help/currency-iso-code-country
https://en.wikipedia.org/wiki/CFA_franc#Central_African
https://en.wikipedia.org/wiki/Central_African_CFA_franc


rposborne:
:ship:

Issue #315, opened 2015-11-29
Country Calling Code


ouranos:
I have a problem with the calling code for some countries. To give a few examples:
* BS (The Bahamas) is listed as 1, whereas it should  be 1242
* BV (Bouvet Island)  is empty and should be 47

Is it me not understanding E.164 or is it an actual issue?

This gem seems to have the correct ones: https://github.com/alexrabarts/iso_country_codes/blob/master/lib/iso_country_codes/calling.rb

rposborne:
@ouranos thanks for pointing this out.  BS (The Bahamas) is technically correct, as they are part of the [NANP](https://en.wikipedia.org/wiki/List_of_North_American_Numbering_Plan_area_codes).  We currently have all other NANP countries setup the same way, +1 as country code, and a 10 digit total number, (therefore including the national area code prefix) 

Where did find the source for Bouvet Island?  as it appears to be uninhabited, with conflicting references to +47 and +55.   This is also amusing http://wikitravel.org/en/Bouvet_Island#Understand .  Apparently there is a weather station on it!

espen:
Bouvet is a Norwegian territory. https://en.wikipedia.org/wiki/Bouvet_Island

+55 is Brazil, where are you finding that reference?

I'm not sure what the phone number to the weather station but as it is Norwegian territory I think +47 should be used. Though I assume the weather station is using a satellite phone when there are visitors there and not a Norwegian phone number.

As a Norwegian I'll put in a imperialistic pull request.

rposborne:
@espen, merged this on the "[Do you have a flag](https://www.youtube.com/watch?v=UTduy7Qkvk8)" perspective. 

For curiosity, Is the weather station occupied? I was under the impression it was not.

espen:
They put up a new weather station there last year because the old was lost to the sea. Some other cabins has also been lost due to strong winds. There are regular scientific expeditions there by the Norwegian Polar Institute but there are no long term inhabitants there. Just a lot of seals.

It's the most isolated island in the world. Apparently first discovered by some French guy.

PR #323, opened 2016-01-08
Added 47 as calling code for Bouvet Island, Norwegian territory

espen:
Sources:
https://en.wikipedia.org/wiki/Bouvet_Island
http://en.advisto.com/country-phone-codes.htm

Fixes one example in #315


rposborne:
:ship:

PR #324, opened 2016-01-12
Palestine submit

samizan:
Translated Palestine into Arabic in Arabic locale translations. Also, removed tanween from some Arab countries (not all have tanween).


rposborne:
:ship:

PR #325, opened 2016-01-21
Romanian Standard VAT rate is 20% as of 2016-01-01

dougal:
See: http://www.vatlive.com/european-news/romania-confirms-vat-rate-cut-to-20-2016/

### DIFF
--- a/lib/countries/data/countries/RO.yaml
+++ b/lib/countries/data/countries/RO.yaml
@@ -38,7 +38,7 @@ RO:
   nationality: Romanian
   eu_member: true
   vat_rates:
-    standard: 24
+    standard: 20
     reduced:
     - 5
     - 9


rposborne:
:ship:

dougal:
The old VAT value is still showing in the 1.2.4 release.

    ~ $ gem list | grep countries
    countries (1.2.4)
    ~ $ irb -r 'countries'
    irb(main):001:0> ISO3166::Country.new('RO').vat_rates
    => {"standard"=>24, "reduced"=>[5, 9], "super_reduced"=>nil, "parking"=>nil}

It looks like it requires a `rake update_cache` and new release to fix?

rposborne:
New gem pushed, Let me know if you are still having issues!

dougal:
Fantastic, thanks.

PR #326, opened 2016-02-08
Make California's 'names' value match other states' style

kwals:
While trying to iterate over each US state, I found that while most states have a single value for the `names` key, California had multiple values. The other states included only the English version of the name, while California included some translations;

```
AZ
{"name"=>"Arizona", "names"=>"Arizona"}
CA
{"name"=>"California", "names"=>["Californie", "Kalifornien"]}
CO
{"name"=>"Colorado", "names"=>"Colorado"}
CT
{"name"=>"Connecticut", "names"=>"Connecticut"}
```

This causes some unexpected behavior when iterating over each state's information. 

If the plan is to add multiple translations for all states, then this PR isn't necessary. However, if there are no plans to do so, I think it would be helpful to keep these consistent.

### DIFF
--- a/lib/countries/data/subdivisions/US.yaml
+++ b/lib/countries/data/subdivisions/US.yaml
@@ -60,9 +60,7 @@ AZ:
   max_longitude: -109.0452231
 CA:
   name: California
-  names:
-  - Californie
-  - Kalifornien
+  names: California
   latitude: 36.778261
   longitude: -119.4179324
   min_latitude: 32.5342321


rposborne:
@kwals thanks for this... our names field is somewhat of a misnomer, names implies it should be an array / hash of some sort but or data shows otherwise.   We do intend to do translations, but it will be on a different key `translations: {en: 'California'...`

rposborne:
@kwals merging this in but expect that names is going to go away around v2 in favor of "unofficial_names" and "translations",  Just a heads up.

kwals:
@rposborne Thanks, looking forward to `unofficial_names` & `translations` :-)

PR #327, opened 2016-02-09
added country name variations for some countries

rvdheijden:
-Democratic Republic of the Congo
-Ivory Coast
-Guinea Bissau
-Palestinian Authority
-St Kitts and Nevis
-St Lucia
-St Vincent Grenadines
-Turks and Caicos

- updated version to 1.2.6


rposborne:
@rvdheijden Just to confirm these are all "slang names" and not official translations?

rvdheijden:
Could call them 'slang names' I suppose, in general they are small variations to the already existing names, eg. no dot after 'St' or no dash between two-part names, or 'and' instead of '&'.
The added names for Congo and Palestina are official names according to the wikipedia.

These small changes were made to work with variety of data we receive from 3rd parties and was submitted back in case you (and others) would find them usefull too.

PR #331, opened 2016-02-19
use only the language of locales like pt-BR to find country local names

cerdiogenes:


rposborne:
@cerdiogenes How would this affect languages, like `zh-Hans` if we were to add them?

cerdiogenes:
@rposborne it will work the same as for pt-BR. The proposed PR uses only the language component of the locale, "pt" for "pt-BR" and will use "zh" for "zh-Hans" and any other locale with "-".

The po files in the debian package [iso_codes](http://anonscm.debian.org/cgit/pkg-isocodes/iso-codes.git/tree/iso_639/zh_CN.po#n1561), used by i18n_data, don't use dialects, like "Portuguese (Brazil)", only the language "Portuguese". I guess that this is a standard.

rposborne:
@cerdiogenes, I am thinking this should be a fallback. I do plan to add dialects and think this is good behavior but it needs to not just default to only `pt` when `pt-BR`  is passed in.

rposborne:
I think stubbing out the translations method for X country, that returns a key that contains a key with the `-` and one that passes in `pt-br` and expects to fallback to `pt`.

rposborne:
@cerdiogenes I had added a test on a new PR for handling the fallback.   Will you look at it to ensure it still meets your requirements.  I will close this PR in favor of #379

cerdiogenes:
With this patch I'm still getting country names in english for `ISO3166::Country.translations('BR')`.

I guess the root of the problem is what I commentend in https://github.com/hexorx/countries/pull/331#issuecomment-190238042: "The po files in the debian package iso_codes, used by i18n_data, don't use dialects, like "Portuguese (Brazil)", only the language "Portuguese". I guess that this is a standard."

I will give it a second look and try to report back this week.

Best regards!

rposborne:
@cerdiogenes You seem to be using that method, incorrectly, `ISO3166::Country.translations('BR')` will attempt to translate all countries to the language `Breton`  vs `ISO3166::Country.translations('pt-BR')` which will attempt to translate all countries to `Brazilian Portuguese`, but currently the PR as it stands does not address accessing the data in this manner only from the `local_names` is the fallback handled.

PR #333, opened 2016-03-01
Update PH Subdivisions

danshultz:
There were 2 subdivisions missing which I have added - DIN and DVO

There is also a set of regions for PH https://en.wikipedia.org/wiki/ISO_3166-2:PH

All subdivisions are within a region except for the Region of `00` which has no subdivisions. I've added `00` and could add the other subdivisions (similar to what appears to exist for the UK). 

Is this PR good or should I add the additional PH regions?


rposborne:
Awesome @danshultz!  Thanks for this, :ship:

Issue #346, opened 2016-05-17
Missing country locale file throws Errno::ENOENT Exception


jstenhouse:
When using version `2.0.0.pre.3` and [I18n.available_locales](https://github.com/hexorx/countries/blob/master/lib/countries/configuration.rb#L31) returns a locale that doesn't have a backing json file, [Data#sync_translations](https://github.com/hexorx/countries/blob/master/lib/countries/data.rb#L62) throws an `Errno::ENOENT Exception: No such file or directory @ rb_sysopen`

Ex.
```
(byebug) ISO3166::Country.find_country_by_alpha2('TR')

*** Errno::ENOENT Exception: No such file or directory @ rb_sysopen - 
/Users/jason/.rbenv/versions/2.3.1
/lib/ruby/gems/2.3.0/gems/countries-2.0.0.pre.3/lib/countries/cache/locales/pt-pt.json
```

The `pt-pt` locale is getting added by another gem [password_strength](https://github.com/fnando/password_strength/blob/master/locales/pt.yml) in my app -- not sure why but I think countries shouldn't throw an exception on an unsupported locale.

It looks like [Data#load_cache](https://github.com/hexorx/countries/blob/master/lib/countries/data.rb#L123-L125) isn't checking for the existence of the file before using `File#binread`. 

And I checked that this doesn't throw an exception for version 1.2.5. 

Want me to open a PR with an added `File#exists?` for `load_cache`?

Thanks!

rposborne:
@jstenhouse a PR would be awesome, through a test in to capture an unexpected locale and you will be my hero!

rposborne:
Merged, new release will be shipped in minutes.

PR #347, opened 2016-05-18
fix Data#load_cache from throwing exception when missing locale file

jstenhouse:
@rposborne 

Fix and test for #346 . 👍


rposborne:
🚢

PR #348, opened 2016-06-03
Explain in README regarding global Country class helper.

giedriusr:
Updated version of README should help avoiding conflicts as I had such a scenario.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -26,6 +26,10 @@ Or you can install via bundler Gemfile with a Country Helper class:
 
     gem 'countries', :require => 'countries/global'
 
+If you already have your own Country class to avoid conflicts you can simply require iso3166:
+
+    gem 'countries', :require => 'countries/iso3166'
+
 Upgrading to 1.2.0
 -----------
 


rposborne:
@giedriusr, not including the global helper is now the default behavior. Maybe it's best explain why the global line exists, so they don't jump to include the global.

rposborne:
@giedriusr This is actually a really good spot, thanks for pointing this out. I didn't realize how misleading the start of the Readme was. 👍

giedriusr:
@rposborne it's never easy to catch all the glitches when you're the creator. Didn't understand your request, should I update this PR with an extra info?

rposborne:
@giedriusr It would be great if you could, points that we need to get across. 

- By default you should not have a require in the gemfile
- If you NEED to use `Country` and risk collision you should require `countries/global`
- Upgrade notes are specific to the global helper

giedriusr:
I see your point and that makes sense to have it like that by default, but the weird thing is actually that whenever I remove require from Gemfile my specs break. So maybe there is a bug somewhere? Or am I doing something wrong :)
```
$ bundle | grep 'countries'
Using countries 1.2.5
```

Switching back to `gem 'countries', '~> 1.2.0', require: 'countries/iso3166'` fixes everything.

rposborne:
@giedriusr I spent some time this morning to try to reproduce this on a new app.  I don't see `Country` being required only `ISO3166::Country`

rposborne:
@giedriusr Can we split this out to an issue?  I have tried to clean up the readme based upon this PR.

giedriusr:
👍

PR #356, opened 2016-07-28
Update .travis.yml

hexorx:
Removed older versions of Ruby from tests.


rposborne:
@hexorx Do we want to fully drop these older rubies. I am 👍 if you are?

hexorx:
Yeah let's. Old ruby can use older gem version.

PR #369, opened 2016-09-14
add vat rates for CH

afdev82:
I added the vat rates for CH.
I didn't understand what super_reduced is used for.


rposborne:
Some EU member states have policies that will reduce tax rates further.

PR #379, opened 2016-09-19
use only the language of locales like pt-BR to find country local names / with fallback

rposborne:
This superseeds #331 with adding of a fallback if a specfic language has a locale specfic translaton


rposborne:
@cerdiogenes You seem to be using that method, incorrectly, `ISO3166::Country.translations('BR')` will attempt to translate all countries to the language `Breton`  vs `ISO3166::Country.translations('pt-BR')` which will attempt to translate all countries to `Brazilian Portuguese`, but currently the PR as it stands does not address accessing the data in this manner only from the `local_names` is the fallback handled.

rposborne:
@cerdiogenes So I have taken a stab at making the entire system to be agnostic to a request of a locale like `pt-BR`, which will fallback to just `pt`

cerdiogenes:
Fantastic @rposborne, thanks for addressing this!

PR #382, opened 2016-09-27
Change Zimbabwe currency code to USD

rlt:
The successor is difficult, but the Zimbabwean Dollar was demonetised in 2015 and replaced with a basket of currencies. The government uses the USD in official transactions, however.


rposborne:
I see your point, and as far as I know we can't handle multiple currencies.   We might want to tag an issue for a manner to track multiple currencies.

PR #386, opened 2016-09-28
Fix typos in specs and readme

contrepoint:
noticed some typos while going through the gem

thanks for making the gem - I use it regularly


rposborne:
Thanks!

PR #392, opened 2016-11-10
added NPN code after country code NANP

maomaoZH:
added 19 countries NPN code


rposborne:
This appears to be breaking previous implementations, but this we do want this data.   Can we add a new field for this code? or for the combination?

maomaoZH:
I can do another pull request for this data.  What field name do you suggest?

rposborne:
@maomaoZH maybe something like `nanp_area_code` or `nanp_prefix`

maomaoZH:
ok. will do it.

Issue #395, opened 2016-11-15
Location for Vatican is now Virginia (US)


IngoAlbers:
It seems that here: https://github.com/hexorx/countries/commit/c72327e1f8223c7fcc65455b9f69e917347b8dd9#diff-cec408aa407cf5acca7ee93cc4ecd98e
the bounding box of Vatican was accidentally set to the one of Virginia the US state.

I stumbled across this by accident while looking at a problem that was caused by the values that were there before. https://github.com/hexorx/countries/blob/v1.2.5/lib/countries/data/countries/VA.yaml#L37

As you can see the min and maximum values for latitude and longitude are identical in which case the actual area will be 0, which might cause problems in some use cases.

rposborne:
@IngoAlbers Yup can confirm. Patch release will be inbound later today.

rposborne:
@IngoAlbers 2.0.1 just got pushed to rubygems.

PR #398, opened 2016-11-16
Improve bounding box lookup

rposborne:
Previous implmentation had issues arround various countries.  We now use alpha2 plus the word "country" for better results.


rposborne:
This should fix #395

PR #403, opened 2016-11-19
Restrict money version to ~> 6.7

camertron:
This gem (countries) depends on money ~>6.0 and requires `sixarm_ruby_unaccent` in countries.rb. The problem is, `sixarm_ruby_unaccent` is a dependency of money v6.7 and up. One of our gems depends on an earlier version of money, so countries >= 2.0 caused our builds to error.

This PR updates countries to depend on money 6.7 and above.

### DIFF
--- a/countries.gemspec
+++ b/countries.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.license       = 'MIT'
 
   gem.add_dependency('i18n_data', '~> 0.7.0')
-  gem.add_dependency('money', '~> 6.0')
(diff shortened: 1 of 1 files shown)


rposborne:
@camertron Awesome, do we need to yank anything? Don't want this to continue to cause errors.

camertron:
Hey @rposborne thanks for the quick merge! Looks like the `require` in question was added in v2.0.0.pre.3, so the 1.2.x series is safe. I don't think there's any need to yank, but I would suggest the next release be at the patch level (i.e. 2.0.2). Thanks again!

PR #412, opened 2016-12-14
Add sixarm_ruby_unaccent to gemspec dependencies

chrisrbnelson:
`countries` requires `sixarm_ruby_unaccent` (as included in `lib/countries.rb`), however it's not listed as a dependency. This causes breakages unless the gem already happens to be installed or it's installed manually.


rposborne:
@chrisrbnelson While I agree this is a correct to add versus us previously implicitly depending on the money gem to bring this in.   I am curious about how the issue came up.

PR #414, opened 2016-12-16
Update all official country names to the ISO 3166 standard

tom-lord:
As discussed here: https://github.com/hexorx/countries/issues/413

### DIFF
--- a/lib/countries/data/countries/BO.yaml
+++ b/lib/countries/data/countries/BO.yaml
@@ -7,7 +7,7 @@ BO:
   international_prefix: '0010'
   ioc: BOL
   gec: BL
-  name: Bolivia
+  name: Bolivia (Plurinational State of)
   national_destination_code_lengths:
   - 2
   national_number_lengths:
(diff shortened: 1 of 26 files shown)


rposborne:
@tom-lord was this done by hand or is there code that was used to generate these?

tom-lord:
I did it with a quick hacky bit of code.... The first step was a little bit manual:

Create a (ruby) hash map of `iso_names = { <country_code> => <country_name>, ...}` based on [this table](https://en.wikipedia.org/wiki/ISO_3166-1). (I did this with a big of text editor manipulation).

Then, I just did the following:

```
iso_names.each do |alpha2, iso_name|
  if ISO3166::Country.find_country_by_alpha2(alpha2).name != iso_name
    `sed -i "s/ name: .*/ name: #{name}/g" "/home/tom/development/countries/lib/countries/data/countries/#{key}.yaml"`
  end
end
```

> I choose a lazy person to do a hard job. Because a lazy person will find an easy way to do it. -- Bill Gates

:laughing:

PR #418, opened 2017-01-09
Fix Norwegian country code (false -> NO)

stefkin:


stefkin:
@rposborne Should I also regenerate countries.json with update_cache and submit it?

PR #421, opened 2017-01-11
Add missing un_locode for south sudan

stefkin:
I've also added a spec to check that each country has a un_locode and regenerated countries cache.


rposborne:
Very good test and all! Thank you for all your hard work!

PR #424, opened 2017-01-15
Adds French Guiana nationality

baron816:
Source: https://en.wikipedia.org/wiki/French_Guiana


rposborne:
@baron816, could you run `rake update_cache` on this.

PR #427, opened 2017-01-24
add Chinese (Traditional) translate json file

showlovel:


rposborne:
@showlovel This is awesome!

But since this is the first Locale based language derivation we are going to need to add some tests to verify that this works as expected.

rposborne:
@showlovel I just noticed that this was made to the cache folder directly which is generated programmatically.  Therefore this file would be overridden.  This pull request cannot be merged in it's current form.

Issue #436, opened 2017-02-20
Corrupt subdivision names


jawnsy:
Many of the subdivision names appear to contain `?` in place of Unicode characters. For example:

"[Aragac?otn](https://github.com/hexorx/countries/blob/master/lib/countries/data/subdivisions/AM.yaml#L13)" should be "[Aragatsotn](https://en.wikipedia.org/wiki/Aragatsotn_Province)"
"[Al ?udaydah](https://github.com/hexorx/countries/blob/master/lib/countries/data/subdivisions/YE.yaml#L124)" should be [Al Hudaydah](https://en.wikipedia.org/wiki/Al_Hudaydah_Governorate)

Are these a problem with some of the tools used to generate this data, or the source data?

jawnsy:
Looks like these alternative sources of data may be better:

* Debian pkg-isocodes, from ISO3166-2 data: https://anonscm.debian.org/cgit/pkg-isocodes/iso-codes.git/tree/data/iso_3166-2.json#n561
* Unicode CLDR: http://www.unicode.org/cldr/charts/latest/supplemental/territory_subdivisions.html#AM
* Official ISO3166-2 data (requires paid subscription): http://www.iso.org/iso/home/standards/country_codes.htm

Could also scrape updated data from Wikipedia or UN/LOCODE, I guess.

PR #438, opened 2017-02-22
Use Subdivision Names from Unicode CLDR

jawnsy:
This change replaces names of subdivisions with those defined in
the Unicode CLDR, if they exist in both places.  It serializes
using Ruby's built-in YAML support, so some keys will be escaped
correctly.

Closes: #436

### DIFF
--- a/lib/countries/data/subdivisions/AD.yaml
+++ b/lib/countries/data/subdivisions/AD.yaml
@@ -80,7 +80,7 @@
     max_longitude: 1.5353385
   name: Andorra la Vella
   comments: 
-08:
+'08':
   unofficial_names:
   - Les Escaldes
   translations:
--- a/lib/countries/data/subdivisions/AE.yaml
+++ b/lib/countries/data/subdivisions/AE.yaml
@@ -10,7 +10,7 @@ AJ:
     min_longitude: 55.4239777
     max_latitude: 25.450004
     max_longitude: 55.6371712
-  name: "'Ajmān"
+  name: Ajman
   comments: 
 AZ:
   unofficial_names:
@@ -29,7 +29,7 @@ AZ:
     min_longitude: 54.268856
     max_latitude: 24.6213301
     max_longitude: 54.8509598
-  name: Abū Z̧aby [Abu Dhabi]
+  name: Abu Dhabi
   comments: 
 DU:
   unofficial_names:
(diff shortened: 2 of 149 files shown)


jawnsy:
Note: These are English names, since those are the only ones provided by the Unicode CLDR files -http://unicode.org/cldr/trac/browser/tags/release-30-0-3/common/subdivisions/en.xml

jawnsy:
This was done programmatically using a script like this (also using subdivision data that has been massaged a bit).

[subdivisions-full.yaml: Full list of subdivisions (including missing)](https://github.com/hexorx/countries/files/795109/subdivisions-full.yaml.txt)

[subdivisions.yaml: List of subdivisions (excluding those missing in countries gem), used with the script](https://github.com/hexorx/countries/files/795108/subdivisions.yaml.txt)

```ruby
#!/usr/bin/ruby

require 'yaml'

subdivisions = YAML.load_file('subdivisions.yaml')

subdivisions.each do |key, value|
  subs = YAML.load_file('subdivisions/' + key + '.yaml')
  s = subdivisions[key]
  subs.each do |key, value|
    if s.key? key
      value['name'] = s[key]
    end
  end

  subs_file = File.open('subdivisions/' + key + '.yaml', 'w')
  subs_file.write subs.to_yaml
end
```

jawnsy:
We will need to carefully review each change on a case-by-case basis, as I'm not sure which of the changes really make sense, nor where the Unicode CLDR sources their data. Ideally we'd update using the latest ISO3166-2 data, but this requires a subscription with the ISO (thus costs 300 CHF per year), and I'm guessing none of us have one.

The names from CLDR are all English translations (with some German available as well), which may or may not be what we want.

rposborne:
This is awesome, and haven't had a chance to fully review, but at first glance currently this overruns the previously localized subdivision names. (Or it appears to at first glance,  check out wien Austria)  I am not sure if this a step in the right direction,  what about adding this as a `translations` key with nested translations so we can begin capturing that data?

jawnsy:
Yes, we can capture these all as English translations, but then we'd still need to manually fix "corrupt" subdivision names like "Aragac?otn" in the linked issue. I'm not opposed to that approach, however. We may also need to be careful due to licensing (I haven't read through the Unicode CLDR license) though I don't think data is considered copyrightable in most countries...

rposborne:
@jawnsy I was able to find more data on the CLDR site. 

http://www.unicode.org/repos/cldr/trunk/common/subdivisions/

rposborne:
@jawnsy I am going to close this out for a by hand correction for the corrupted subdivisions.   Thank you for all your hard efforts on this. 
❤️😍🎉😍❤️🎉😍❤️👏👏👏:

PR #446, opened 2017-03-05
Normalized Turkey’s subdivisions

emir:


jnbn:
Happy to see you've changed it @emir  +1

rposborne:
@emir Thank you so much for the data correction.   Always appreciated!

PR #447, opened 2017-03-09
Write out dependence of active_support in all_names_with_codes

subfusc:
Currently you are left with the following if you 

```
$ gem install countries 
$ irb
irb(main):001:0> require 'countries/global'
=> true
irb(main):003:0> ISO3166::Country.all_names_with_codes
NoMethodError: undefined method 'html_safe' for "Andorra":String
```

The proposed patch writes out the dependency for active_support by removing the use of html_safe with backwards compatibility, although I am unsure it was needed in the first place.

Hope it can be considered as not everyone use active_support all the time. 

Regards
Subfusc

### DIFF
--- a/lib/countries/country/class_methods.rb
+++ b/lib/countries/country/class_methods.rb
@@ -35,7 +35,10 @@ module ISO3166
     end
 
     def all_names_with_codes(locale = 'en')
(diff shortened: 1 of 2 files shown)


rposborne:
Looks good.

PR #450, opened 2017-03-31
Support older rubies

pezholio:
https://rubygems.org/gems/countries/versions/2.0.8 seems to suggest that the latest version of the gems works with any Ruby, however, when trying to install on anything lower than 2.0.0 *, the UTF-8 characters in emojis.rb caused a `invalid multibyte char (US-ASCII)` as older rubies don't autodetect the encoding.

This is easily fixed by adding a magic comment to the file in question, and the tests now happily pass on 1.9.3. I've also had to add a slight tweak the the gemfile, as the VERSION constant is a frozen string, but this doesn't affect the functionality of the gem.

For good measure, I've added `1.9.3` to the Travis setup just to make doubly sure everything passes ongoing.

\* I know 1.9.x is EOL, but we're on legacy infrastructure at the moment, so need to struggle along, at least in the short to medium term.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 cache: bundler
 language: ruby
 rvm:
+ - 1.9.3
  - 2.2.6
  - 2.3.3
  - 2.4.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,3 +13,4 @@ before_install:
 matrix:
   allow_failures:
     - rvm: ruby-head
+    - rvm: 1.9.3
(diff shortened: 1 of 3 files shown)


rposborne:
@pezholio Sorry you are still on 1.9.3. I am happy to accept this change, but I think it's best that we mark 1.9.3 as an allowed failure as we have already stated that it is no longer supported by this gem.

pezholio:
Cool, I've added 1.9.3 to the `allow_failures` matrix 👍

pezholio:
Lovely, thanks 😄

PR #452, opened 2017-04-05
Update KG.yaml

newstler:
Add Kyrgyz

### DIFF
--- a/lib/countries/data/countries/KG.yaml
+++ b/lib/countries/data/countries/KG.yaml
@@ -28,6 +28,7 @@ KG:
   - キルギス
   - Kirgizië
   - Kyrgzstan
+  - Kyrgyz
   languages_official:
   - ky
   - ru


rposborne:
@newstler Is `Kyrgyz` an unofficial or a slang name?  Kyrgyz seems to be the official name for the country?  Is this data incorrect in our translations?

newstler:
Well, I'm not sure, but one of the APIs I use (British one) returns "Kyrgyz" instead of "Kyrgyzstan" so  `ISO3166::Country.find_country_by_name("Kyrgyz")` returns `nil`. So I thought this update may resolve this.

newstler:
Actually I just found that this code doesn't work anyway. How do I associate `Kyrgyz` with `KG` so `ISO3166::Country.find_country_by_name("Kyrgyz")` would return it?

pmor:
@rposborne
The official name is "The Kyrgyz Republic", while "Kyrgystan" is the ISO "short name".(https://www.iso.org/obp/ui/#iso:code:3166:KG)

I have no idea if "Kyrgyz" alone is a commonly used unofficial name, which is what's been added in this PR, so not sure if I should accept or close.

PR #459, opened 2017-07-06
fix inconsistencies in language for BR.yaml

suryast:


suryast:
This issue causes edge case if the gem is used in conjunction with other gems like 'countries'. Also if compared with other countries the entry 'pt-BR' seems inconsistent. The entry should be 'pt' instead?

PR #460, opened 2017-07-13
allow to register subdivisions

kasparsj:
At present subdivisions are always loaded from files (even if the country is custom registered). This fix enables passing subdivison data to `ISO3166::Data.register`


kasparsj:
sure, will do

kasparsj:
@rposborne should be good now

PR #465, opened 2017-08-03
Fix incorrectly named subdivision data files

skyborn8:
Subdivisions for Cayman Islands and Tajikistan is working inconsistently because:
1. On Mac OS file names is case insensitive and `ISO3166::Country['KY'].subdivisions` returning data from file
2. On Linux file names is case sensitive and same code will return empty hash, because library is looking for `KY.yaml` file, but it is named `Ky.yaml`.


rposborne:
@skyborn8 Thanks for this!

PR #473, opened 2017-08-15
Add CLDR data source for subdivion translations.

### LINKED ISSUE #437: Some countries are missing subdivisions
Some countries have subdivisions that are not reported by the countries gem, because the corresponding data file is missing.

* [GL (Greenland)](https://en.wikipedia.org/wiki/ISO_3166-2:GL)
* [LC (Santa Lucia)](https://en.wikipedia.org/wiki/ISO_3166-2:LC)
* [MC (Monaco)](https://en.wikipedia.org/wiki/ISO_3166-2:MC)
* [PS (State of Palestine)](https://en.wikipedia.org/wiki/ISO_3166-2:PS)
* [SS (South Sudan)](https://en.wikipedia.org/wiki/ISO_3166-2:SS)
* [TJ (Tajikistan)](https://en.wikipedia.org/wiki/ISO_3166-2:TJ)
* [WF (Wallis and Futuna)](https://en.wikipedia.org/wiki/ISO_3166-2:TJ)

I found these by looking at the [list of subdivisions defined by the CLDR dataset](http://www.unicode.org/cldr/charts/30/supplemental/territory_subdivisions.html) and noting the ones where no corresponding `data/subdivisions/xx.yaml` file exists.

rposborne:
This massive improves the number of translations we have available and lays the groundwork of other importers.

Thanks to @coderifous and @mig for the help on this.

This fixes #437


jawnsy:
I perused these changes quickly and they look good to me. I think it's a better approach than what I proposed with #438 because it handles updates better (regrettably I didn't understand the way updates are currently handled and couldn't do as good a job as you did there).

Unfortunately I can't test the changes as I left Red Hat and no longer have access to the source code of the project I was working on.

One thing I noticed is that many fields where the data is presumably unknown are recorded as empty fields, and I wonder if it would be better to exclude them from the file entirely? For example, in AM.yaml:

```yaml
  geo:
    latitude: 
    longitude: 
    min_latitude: 
    min_longitude: 
    max_latitude: 
    max_longitude: 
  name: Gegark'unik'
```

It seems the whole `geo` section can be excluded? Over all of the input files this might save some disk space, though probably not a huge amount, and it doesn't matter much with compression.

jawnsy:
@rposborne Sounds like a reasonable approach to me! Thanks for your work!

rposborne:
@jawnsy  2.1.3 released which includes this change

PR #474, opened 2017-08-15
Correcting typos for UY regions

fpinvidio:
Correcting typos for few UY regions names


rposborne:
@fpinvidio Thanks!!

Though it looks like this will conflict with or region overhaul that is about to land #473.

fpinvidio:
I see that. Closing PR .

PR #476, opened 2017-08-21
Add standard vat rate for Norway

### LINKED ISSUE #469: Missing VAT rates
This is more a question than an issue:

Some countries are missing VAT rate information. Is there a reason for that? Would you merge a PR that set vat rate for eg. Norway?

jgrau:
Fixes #469


rposborne:
@jgrau  2.1.3 released which includes this change

PR #478, opened 2017-10-05
Add vat rates for mexico

jgrau:


rposborne:
@jgrau  2.1.3 released which includes this change

Issue #431, opened 2017-01-27
Have a iso_name method and return the common name as name


iGEL:
The issue existed before, but became worse with #414. Some country names are just so ridiculous long that no one would use them in real life (e.g. `"United Kingdom of Great Britain and Northern Ireland"`). While I think it's a good thing to have the ISO names, this gem should also provide the name that 99% of the people use in the daily life. A good hint could be the name of the country's Wikipedia article (in that case: [United Kingdom](https://en.wikipedia.org/wiki/United_Kingdom)).

My suggestion:
```ruby
c = Country["UK"]
c.name # => "United Kingdom"
c.iso_name # => "United Kingdom of Great Britain and Northern Ireland"
```
Obviously it sucks to change the output of `name` again, but I think, in the longer run this is the most intuitive way.

rposborne:
@iGEL I think we have this covered via our translations, which are much much more "common".

rlt:
Chipping in here, but the effect of this change has been noticeable across our apps. I do not think that saying just use the translations is fair. It will be quite a task for us to go and update everything to pull out the short name out from a translations hash. I think @iGEL's suggestion is perfectly fair and reasonable.

rposborne:
Hm I see your point but I am more than happy to change the name of `name` to `iso_name` but I don't think it's a good idea to add something like `name` back in as it becomes WHICH name from WHICH perspective.  And translations are the CORRECT way this problem should be solved.  

That being said, do we need to make the translations API more approachable?   I could see a method like `name` just call out to i18n to get the current locale and load that translation?  With a fall back to a default locale of `en` if we are not inside of rails?

perspectivezoom:
From a usage perspective, I would benefit from having a single method to get the common country name that users expect.

- `ISO3166::Country.all.map(&:name)` gives me the extremely long names that @iGEL mentioned, with the UK name being 52 characters long
- `ISO3166::Country.all.map { |c| c.translations['en'] }` gives me "Viet Nam", which is the official way of doing it, but my users don't want that in a dropdown. It also includes "Korea, Republic of" and "Korea, Democratic People's Republic of". I can't expect my users to know the difference.
- `ISO3166::Country.all.map { |c| c.unofficial_names[0] }` is definitely the closest, with only a small number of unexpected names. For some reason `KP` is "Korea (North)", whereas `KR` is "South Korea".

[...]

PR #485, opened 2017-10-23
Fix incorrect name to TW

WilliamLINWEN:
Fix incorrect name to TW


rposborne:
This is generally a contested change. And I have avoided accepting PR's involving Taiwan's name. I am in no way able to make this choice on my own.

lib/countries/data/countries/TW.yaml
```diff
@@ -12,7 +12,7 @@ TW:
   international_prefix: '002'
   ioc: TPE
   gec: TW
-  name: Taiwan, Province of China
+  name: Taiwan, Republic of china
```
jlnr: *China should be capitalized even here

jlnr:
Some arguments for changing this entry:

* All the big players in IT call Taiwan just "Taiwan"; see e.g. [Google's region picker](https://store.google.com/de/countrypicker?hl=en-US), the macOS/iOS Region preferences, the country picker on dell.com, ...
* Even in [the recent series of back-and-forth about how Taiwan is called on websites](https://www.taiwannews.com.tw/en/news/3382136), websites flipped between "Taiwan" and "Taiwan, China" due to political pressure. No company seems to follow the ISO-3166 list.
* Postal mail sent to Taiwan is conventionally sent to "Taiwan, ROC" because otherwise [it can disappear](https://sentinel.tw/mail-not-arriving-taiwan/) – not something you want to have happen to users.

I think the right approach would be to include the unmodified ISO-3166 name, but *also* a property that identifies how countries are conventionally referred to on public-facing websites. People who use this gem through the `country_select` gem, for example, are not primarily looking for an ISO standard, but for an uncontroversial and unambiguous list of countries that will work on a website.

jlnr:
OK, so later today I realized that the reason most companies list Taiwan as "Taiwan" is that they use the Unicode CLDR data, not ISO-3166. There's some good discussion about this in this [Drupal issue](https://www.drupal.org/project/drupal/issues/1938892).

Then I also noticed that #431 already mentions the `twitter-cldr` gem 🎉

pmor:
As per https://www.iso.org/obp/ui/#iso:code:3166:TW, the official ISO 3166 name is "Taiwan, Province of China", so I'm closing this, same as I did #601. See my comment there for more information, but keep in mind the gem provides "common" names via the `translations`

PR #488, opened 2017-12-13
Update BY currency code (BYR -> BYN)

mainameiz:
https://en.wikipedia.org/wiki/Belarusian_ruble

> In July 2016, a new ruble was introduced (ISO 4217 code BYN), at a rate of 1 BYN = 10,000 BYR.


rposborne:
Thanks!!!

rposborne:
@mainameiz  2.1.3 released which includes this change

PR #490, opened 2017-12-22
Fix search to work for countries with comma in their name

budmc29:
Using ISO3166::Country.find_country_by_name('Korea, Republic of')
returns nil, since Korea is stored as "Korea (Republic of)".

This change will normalize all the terms that are used to search and
those in the Data.cache.

### DIFF
--- a/lib/countries/country/class_methods.rb
+++ b/lib/countries/country/class_methods.rb
@@ -1,6 +1,6 @@
 module ISO3166
   UNSEARCHABLE_METHODS = [:translations].freeze
-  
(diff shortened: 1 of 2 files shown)


budmc29:
Thank you very much.

rposborne:
Welcome, I'll try to cut a release soon.

Ping me again if it's not don by next friday :)

rposborne:
@budmc29 2.1.3 released which includes this change

budmc29:
That's great!

PR #500, opened 2018-02-22
🇬🇧🇪🇺 Britain leaves the EU and EEA

seanhandley:
For when the day actually comes 😢

### DIFF
--- a/lib/countries/data/countries/GB.yaml
+++ b/lib/countries/data/countries/GB.yaml
@@ -27,8 +27,8 @@ GB:
   world_region: EMEA
   un_locode: GB
   nationality: British
-  eu_member: true
-  eea_member: true
+  eu_member: false
+  eea_member: false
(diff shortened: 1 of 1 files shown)


rposborne:
@seanhandley Is there a day that this is supposed to happen? (I can't tolerate looking at the news these days)

seanhandley:
March 29th 2019 is the absolute limit for all agreements to be finalised. Even if they're not finalised by then, by the terms of Article 50 Britain will officially be out of the union and the EEA.

pmor:
Looks like January 31st 2020 it is.

The Brexit agreement was signed a few hours ago and the final vote in the European Parliament  should be on January 29th.

seanhandley:
Wow, nearly two years later and it's time to merge this.

What an agonising and upsetting time it's been. So long, EU 💔

PR #505, opened 2018-03-02
Corrected swedish transation for GB

pidu:
The former correction was not made under the swedish locale but country code.


pidu:
Am I correctly assuming I should not be running the rake task myself?

rposborne:
@pidu you certainly can run the rake task.  It makes it easier for me to merge.  Though this seems to be throwing a faliure.

pidu:
Were you refering to the build failure, I thought that was a fluke?

PR #510, opened 2018-03-15
Memoize subdivision YAML loading

mdehoog:
Currently the subdivision YAML files are being loaded for every new `Country` instance. This PR memoizes the loaded YAML file on a class instance variable.

Also replaced the `inject` -> `merge` pattern with a more performant version.

Performance changes:
From:
```ruby
pry(main)> Benchmark.realtime { 10000.times { Country['US'].subdivisions } }
=> 17.20118099998217
```
To:
```ruby
pry(main)> Benchmark.realtime { 10000.times { Country['US'].subdivisions } }
=> 0.6878289999440312
```

### DIFF
--- a/lib/countries/country.rb
+++ b/lib/countries/country.rb
@@ -56,13 +56,15 @@ module ISO3166
     end
 
     def subdivisions?
-      !data['subdivisions'].nil? || File.exist?(subdivision_file_path)
+      !subdivisions.empty?
     end
 
     def subdivisions
-      @subdivisions ||= subdivision_data.inject({}) do |hash, (k, v)|
-        hash.merge(k => Subdivision.new(v))
-      end
+      @subdivisions ||= if data['subdivisions']
+                          self.class.create_subdivisions(data['subdivisions'])
+                        else
+                          self.class.subdivisions(alpha2)
+                        end
     end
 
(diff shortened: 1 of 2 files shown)


rposborne:
@mdehoog What are the memory benchmarks on this?  But this is awesome.

mdehoog:
Before:
```ruby
pry(main)> Benchmark.memory { |x| x.report { Country.all.each(&:subdivisions) } }
Calculating -------------------------------------
                        16.663M memsize (     0.000  retained)
                       130.895k objects (     0.000  retained)
                        50.000  strings (     0.000  retained)
```
After:
```ruby
pry(main)> Benchmark.memory { |x| x.report { Country.all.each(&:subdivisions) } }
Calculating -------------------------------------
                         8.122M memsize (     4.710M retained)
                       121.002k objects (    64.401k retained)
                        50.000  strings (    50.000  retained)
```

More memory is retained (~4.7mb), but interestingly not as much is allocated.

rposborne:
And for clarity these only loaded if you call subdivisions?  So no performance change if you are not using subdivisions?

mdehoog:
Correct, the subdivisions are still loaded lazily, but are now memoized at a class level rather than an instance level.

rposborne:
Well hell yea, this is great.

PR #548, opened 2018-10-23
I18ndata for translations

isaiah:
Instead of maintaining the locale files, we can just use the translations from `I18nData` which is already a dependency. Most importantly, `I18nData` is more complete, it contains translations such as traditional Chinese.

### DIFF
--- a/countries.gemspec
+++ b/countries.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.license       = 'MIT'
 
   gem.add_dependency('i18n_data', '~> 0.8.0')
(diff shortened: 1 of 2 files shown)


rposborne:
We have actually done this in the past and it has a major issue.  It forces an in memory reload on every call causing a slow down when trying to generate, a country list (our prime use case). We already technically pull this data just during a prerelease process vs during run time

Check out our rake task over here https://github.com/hexorx/countries/blob/master/Rakefile#L57-L90

isaiah:
@rposborne Thanks for the reply. Could you tell me how do I add traditional Chinese translations without touching the gem?

PR #549, opened 2018-10-31
Update cache

jgrau:
I didn't realize there was this cache file that needs updating when a new datapoint is added. I should do that whenever I add data before making a PR, right?


rposborne:
Yup.

PR #565, opened 2019-01-09
It should include language variants which are available in i18n_data.

torumori:
The current cache data is missing the language variants which are available in i18n_data. Particularly for Chinese, it's critical to include zh_TW and zh_HK translations.

- i18n_data package also contains 'zh_CN', 'zh_HK', 'zh_TW', 'bn_IN', and 'pt_BR' (these are added in [here](https://github.com/grosser/i18n_data/blob/master/lib/i18n_data/file_data_provider.rb#L16)
- 'zh_CN' should be the same as 'zh' since ['zh' is generated from 'zh_CN' translations in i18n_data](https://github.com/grosser/i18n_data/blob/master/lib/i18n_data.rb#L57) This is the reason that `translation_corrections.yaml` needs a correction for zh_tw as well. The same relationship applies to 'bn' and 'bn_IN'


misty44:
when will it merged to master ? I need zh_TW

rposborne:
@torumori Are there any tests that we could add for this?

torumori:
@rposborne I wondered what kind of test I should add, but I feel that to write a test based on the current translation won't add much of values. In the end, this is all about translations data, and I observed that there seem to be not many tests are added along with the cache data changes in the past. That's why I didn't add any test in this case.
If you think differently, I could add something; please let me know.

torumori:
@rposborne I've added the unit tests, please consider merging this PR. Thank you!

tilsammans:
@hexorx is this good to go? I would also very much like to update i18n_data as well: #572

rposborne:
Great addition.

PR #570, opened 2019-01-26
Add North Macedonia as alternate name

toddwschneider:
Macedonia will soon officially be known as the Republic of North Macedonia, per https://www.bbc.com/news/world-europe-47002865

North Macedonia is an appropriate alternate name effective immediately


toncid:
As of 12 Feb, this is now the official name of the country: https://en.wikipedia.org/wiki/North_Macedonia

PR #571, opened 2019-01-28
remove bin/console from gemspec

patleb:
bin/console shouldn't be packaged as an executable


davidstosik:
Thanks for this! ❤️ 
Will we get a new release soon? 🙏 

```
wisper's executable "console" conflicts with countries
Overwrite the executable? [yN]  N
ERROR:  Error installing cp8_cli:
        "console" from wisper conflicts with installed executable from countries                                                                              
```

PR #574, opened 2019-03-12
Fix state codes from being returned as false

akiellor:
This bug was caused by YAML being "helpful" and parsing "NO" as false.
YAML will also parse "YES", "No", "no", "on", "off" and many other
variations as boolean.

https://makandracards.com/makandra/24809-yaml-keys-like-yes-or-no-evaluate-to-true-and-false

### DIFF
--- a/lib/countries/data/subdivisions/FO.yaml
+++ b/lib/countries/data/subdivisions/FO.yaml
@@ -1,5 +1,5 @@
 ---
-NO:
+"NO":
   unofficial_names: Nordoyar
   translations:
     en: Nordoyar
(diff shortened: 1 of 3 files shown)


akiellor:
@rposborne Any thoughts on this one?

lrworth:
This PR is necessary!

In 3.0.0 we have
```
irb(main):010:0> ISO3166::Country['FO'].states['NO']
=> nil
irb(main):011:0> ISO3166::Country['FO'].states[false]
=> #<struct ISO3166::Subdivision name="Nordoyar", code=nil, unofficial_names="Nordoyar", geo={"latitude"=>62.28333, "longitude"=>-6.58333, "min_latitude"=>nil, "min_longitude"=>nil, "max_latitude"=>nil, "max_longitude"=>nil}, translations={"en"=>"Nordoyar"}, comments=nil>
```

rposborne:
Yup solid fix.  Thanks.

PR #576, opened 2019-03-22
Remove the `unicode_utils` gem dependency

torumori:
Given that [Ruby 2.3 EOL is scheduled for 2019-03-31](https://www.ruby-lang.org/en/downloads/branches/), I propose that we should remove the `unicode_utils` gem dependency since it's just used only for `UnicodeUtils.downcase` method. Instead, `String#downcase` should be able to replace it.

- The String#downcase supports non-ASCII from Ruby 2.4.0
- `unicode_utils` gem is no longer maintained, we should rely on more on Ruby core

### DIFF
--- a/countries.gemspec
+++ b/countries.gemspec
@@ -16,7 +16,6 @@ Gem::Specification.new do |gem|
   gem.license       = 'MIT'
 
   gem.add_dependency('i18n_data', '~> 0.10.0')
-  gem.add_dependency('unicode_utils', '~> 1.4')
   gem.add_dependency('sixarm_ruby_unaccent', '~> 1.1')
   gem.add_development_dependency('rspec', '>= 3')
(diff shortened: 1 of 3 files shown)


torumori:
The failure in Ruby 1.9 is caused by system error.
The failed test is expected in Ruby 2.3. It should fail in Ruby 1.9 as well even without a system error.

mauro-oto:
@torumori @rposborne shouldn't Ruby 2.3 be moved to the allowed failures section now that it's past EOL? Or even drop support for Ruby 1.9 and 2.3. I'd like to get rid of this dependency on an app, and this gem is the only blocker left. Thanks!

rposborne:
I support dropping 2.3.  Lets make the required adjustments to get this to pass. @torumori

PR #581, opened 2019-05-14
Changes Norway translation into word used by 95% of its popullation

bytnar:
Changes Norway translation into word used by 95% of its popullation according to wiki
_Bokmål is an official written standard for the Norwegian language, alongside Nynorsk. Bokmål is the preferred written standard of Norwegian for 85% to 90% of the population_


pmor:
Looking into ISO data, it looks like "Norge" is used in Bokmål (locale `nb`) and "Noreg" in Nynorsk (locale `nn`).

If Bokmål is preferred, then it's already using "Norge" as expected. 
The proposed change is to change the Nynorsk translation to use the Bokmål version of the word, and that contradicts both https://www.iso.org/obp/ui/#iso:code:3166:NO and the Unicode CLDR data we use for the localised names.

PR #597, opened 2019-10-02
removes extra space from "UM" Russian locale

dimanyc:
`Соединенные штаты   Малых Удаленных островов` to `Соединенные штаты Малых Удаленных островов`


rposborne:
Is this a hand edit of the cache file? If so it will almost be certainly deleted by accident. If this file needs correcting please add an entry to https://github.com/hexorx/countries/blob/master/lib/countries/data/translation_corrections.yaml

PR #601, opened 2019-10-08
Taiwan is not China

Lewiscowles1986:
We keep having this come up at work. It looks like there was already a PR for this, but just in-case (it would reduce dev friction our side not to maintain a fork just to avoid this)


pmor:
As per https://www.iso.org/obp/ui/#iso:code:3166:TW, the official ISO 3166 name is "Taiwan, Province of China", so I'm closing this.
Taiwan's status is a political matter, and this gem keeps out of those by following the ISO 3166 data.

The unofficial names include "Taiwan", and the translations seems to also use it instead of the ISO. Personally, I've adopted the use of the "common" names via the `translations` attribute over the official names due to issues such as this one, and in particular,  the UK, officially "United Kingdom of Great Britain and Northern Ireland".

The gem's use of of both ISO-official and unofficial names is a little messy right now, granted, and could probably use some clarification, for instance with `iso_name` and `common_name` methods, or something to that effect.

jlnr:
> Taiwan's status is a political matter, and this gem keeps out of those by following the ISO 3166 data.

No, it's the other way around: ISO 3166 is a politically motivated list, Unicode CLDR is the dataset that tries to sidestep political issues. I have linked to related discussions in #485. By using ISO 3166, this library takes a clear political stance.

PR #602, opened 2019-10-10
Fix a typo in 'Portugal'

### LINKED ISSUE #583: 'Portugal' is misspelled in lib/countries/cache/locales/nn.json
It is spelled "Portugual" instead of "Portugal"

mbirman:
Portugual -> Portugal. Closes #583


rposborne:
Is this a hand edit of the cache file? If so it will almost be certainly deleted by accident.  If this file needs correcting please add an entry to https://github.com/hexorx/countries/blob/master/lib/countries/data/translation_corrections.yaml

mbirman:
Hey @rposborne, thanks for the tip. I've added the correction. The comment in that file suggests filling a bug report to http://alioth.debian.org/projects/pkg-iso-codes which seems to be unavailable at the moment. I'm going to try again a bit later.

rposborne:
@mbirman hmm that part of the readme is old,  the repo have moved.  https://salsa.debian.org/iso-codes-team/iso-codes might be it. 

The data flows from debian, to i18n_data, to countries.  so it's possible that our cache is wrong and i18n_data has an update for us.

pmor:
There's an open issue (#607) to update i18n_data, I suggest handling that and validating if this fix is still needed afterwards.

PR #608, opened 2019-12-17
Potential solution for thread safety

hammerdr:
I wouldn't call this PR ready until someone can confirm that this is a real fix.

I attempted to use the small reproduction of the thread safety issue but could not get it to fail locally.

In general, though, this puts a semaphore around access to the `ISO3166::Data.config`, ensuring that threads aren't entering this method during another thread's execution of this.

### DIFF
--- a/lib/countries/data.rb
+++ b/lib/countries/data.rb
@@ -5,6 +5,7 @@ module ISO3166
     @@cache_dir = [File.dirname(__FILE__), 'cache']
     @@cache = {}
     @@registered_data = {}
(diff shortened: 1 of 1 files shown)


rposborne:
Hmm, is there anyway we can add a test for this?  I guess it really can't hurt.  I have never been able to pin down the thread issue myself.  

@hammerdr do you for see any drawbacks to adding the mutex here?

erkie:
I haven't received a deadlock since i applied this fix. Can it be merged now?

PR #611, opened 2020-01-30
Changed name to `North Macedonia`

ademdc:
Even though the new name Macedonia is changed in `lib/countries/data/countries/MK.yaml` which was included in the new version, the old name is present in `lib⁩/countries⁩/cache⁩` which is loaded on a `ISO3166::Country.all` call. 

So I changed the name of `Macedonia (the former Yugoslav Republic of)"` to `North Macedonia` in `lib⁩/countries⁩/cache⁩`


ademdc:
Because the change that I did was fixed in some of the previous commits, I am closing this pull request.

PR #618, opened 2020-02-12
Add data & script to update postal code & format

### LINKED ISSUE #366: Include country's postcode/zip format
I'd like a new field for each country to store the valid postal code / zip code syntax as a regular expression.

This would complement the `postal_code` (boolean) flag in allowing us to validate a country's postcode, for example.

The regex for `GB` postcodes is a monster but all other countries' postcodes seem quite simple.

A good data source is [CLDR](http://cldr.unicode.org/index/downloads) (via [StackOverflow](http://stackoverflow.com/questions/578406/what-is-the-ultimate-postal-code-and-zip-regex/7185241#7185241)).

For `US` (for example) the value would be `postal_code_format: "\d{5}([ \-]\d{4})?"`.

Would this be out of scope or should I provide a PR?

cover:
This adds a script that fetches postal codes data from Google's [Address Data Service](https://chromium-i18n.appspot.com/ssl-address/data) ([doc](https://github.com/google/libaddressinput)), which is used in Android and Chromium.

It sets the `postal_code` if the country has it, and also adds a new `postal_code_format` with the regex returned by Google.

I've decided to just return the format as a string because I wanted to avoid having to initialize a new Regexp every time (especially when returning all countries), as someone may just want to use them as strings (our case), and it should be easy enough to do a `Regexp.new(country.postal_code_format)` when needed.

The base url is set as `https://chromium-i18n.appspot.com/ssl-address/data/`, interestingly, most of the countries are redirected to another url (eg. [US](https://www.gstatic.com/chrome/autofill/libaddressinput/chromium-i18n/ssl-address/data/US)), while 3 ([KP](https://chromium-i18n.appspot.com/ssl-address/data/KP), [MY](https://chromium-i18n.appspot.com/ssl-address/data/MY), and [SY](https://chromium-i18n.appspot.com/ssl-address/data/SY)) are not and return 404 on the gstatic domain :man_shrugging::man_shrugging::man_shrugging: Reason why I've used `open-uri` instead of `net/http` (the former follows redirects without any extra code)

I've also added an alias to support `zip_format`, updated the tests and refreshed the cache to have them passing.

It also updates some format in the yaml since they're been rewritten from Ruby, let me know if it's fine, otherwise I can just add the postal codes related lines and skip the others.

Some countries that were set having a `postal_code` (eg. AQ) are now set to not support it anymore, as it's considering Google as the source of truth.

It's also safe to re-run the script in the future as it'll take care of updating the data.

Closes #366


rposborne:
Looks great.

PR #623, opened 2020-03-20
fix armenian translation

MateoGreil:
linked to this issue : https://github.com/hexorx/countries/issues/622


MateoGreil:
There is no error in your library. Just see the issue.

PR #625, opened 2020-04-06
Fix ISO3166::Country respond_to_missing?

biinari:
Fix `ISO3166::Country.respond_to_missing?` for finder methods like `find_country_by_name`

## Reason

Using rspec-mocks to stub methods on `ISO3166::Country` is broken. As an example, the following snippet raises an error:

```ruby
let(:found_country) { double }
before
  allow(ISO3166::Country).to receive(:find_country_by_alpha2)
    .and_return(found_country)
end
```

Raises:

```
NoMethodError:
  undefined method `all?' for "alpha2":String
```


dima4p:
Yes, please. I need it too

PR #626, opened 2020-04-14
Add country code 672 to HM

wongyouth:
Refer to https://www.iso.org/obp/ui/#search
'Heard Island and McDonald Islands' get a country code of 334

**UPDATE**: actually, the country code is 672.

### DIFF
--- a/lib/countries/data/countries/HM.yaml
+++ b/lib/countries/data/countries/HM.yaml
@@ -3,7 +3,7 @@ HM:
   continent: Antarctica
   alpha2: HM
   alpha3: HMD
-  country_code: ''
+  country_code: '334'
   international_prefix: ''
   ioc: 
   gec: HM
--- a/lib/countries/data/countries/HM.yaml
+++ b/lib/countries/data/countries/HM.yaml
@@ -11,7 +11,7 @@ HM:
   national_destination_code_lengths: []
   national_number_lengths: []
   national_prefix: ''
-  number: '334'
+  number: '672'
   region: ''
   subregion: ''
   un_locode: HM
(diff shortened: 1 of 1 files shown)


pmor:
@rposborne

siklodi-mariusz:
@wongyouth I'm also having issues with my project because of the missing country code of _Heard Island and McDonald Islands_. Unfortunately, the code you added is not the correct one. 334 is referring to the assigned country number and it can be found here:
https://github.com/hexorx/countries/blob/master/lib/countries/data/countries/HM.yaml#L14

The `country_code` yaml key is actually referring to the ISD (International Subscriber Dialing) phone country code. I checked multiple countries and it seems that this is the case. For example, I live in Romania, and our code is 40 and this is what Romania has set as `country_code` here:
https://github.com/hexorx/countries/blob/master/lib/countries/data/countries/RO.yaml#L11

So in the case of _Heard Island and McDonald Islands_ the ISD seems to be 672.
Here are a few sources I found online about this:
https://www.goldenchennai.com/isd-code/isd-code-for-heard-island-and-mcdonald-islands/
https://46elks.com/kb/country-codes
https://en.wikipedia.org/wiki/List_of_country_calling_codes  (The good old Wikipedia 😄 )

So do you think you could make this change?

siklodi-mariusz:
Also for anyone stumbling upon this and looking for a quick fix. 

This is what I did in my project (Rails):

```ruby
# config/initializers/countries.rb

ISO3166::Data.register(
  ISO3166::Country.find_country_by_alpha2('HM').data.with_indifferent_access.merge(country_code: '672')
)

```

wongyouth:
Thanks @siklodi-mariusz, you are right. I'm glad to change this.

PR #628, opened 2020-04-30
especify that there are no executables in gemfile

vaporyhumo:
This addresses https://github.com/hexorx/countries/issues/627

Adding this will remove `bin/console` from the built gem, since it's not meant to be used outside of it, thus preventing name clashes with other gems.


vaporyhumo:
I guess there have been no updates on permissions? @pmor @rposborne @hexorx

vaporyhumo:
Any decisions on this one :wave: ?

PR #635, opened 2020-06-30
Used the common name República Checa in PT

basex:
República Checa is the common name used in Portuguese for Czech Republic


pmor:
@rposborne

PR #644, opened 2020-08-20
Drop support for Ruby < 2.4.0 and remove `unicode_utils`

mauro-oto:
This PR is a follow up to https://github.com/hexorx/countries/pull/576, which intended to replace `unicode_utils` with the native `String#downcase`, since it supports non-ASCII from Ruby 2.4.0. 

I added a commit to it so that we also drop support for Ruby versions < 2.4.0 as we discussed in that PR, which might warrant a major version release to avoid breaking other projects. 

Thanks @torumori and @rposborne!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 cache: bundler
 language: ruby
 rvm:
- - 1.9.3
- - 2.3.6
- - 2.4.3
- - 2.5.0
+ - 2.4.10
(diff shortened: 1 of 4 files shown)


pmor:
These changes were merged into #650 which will supersede this PR, and updated to the current point versions of 2.5, 2.6 and 2.7, and the addition of removal of support for ruby 2.4

PR #649, opened 2020-10-18
Add region names

aviemet:
Adds a property on countries called `subdivision_name` which is the noun used to describe the territory_1/territory_2 regions which should appear on addresses. Names were taken in part from [this wikipedia article](https://en.wikipedia.org/wiki/Matrix_of_country_subdivisions) with some additional research.


rposborne:
This looks great, though question should this be a translated field?

pmor:
@rposborne Great question. 
Given that ISO country names are in English (at least most of them, AFAIK) and then we have translations (eg "Spain" vs "España"), I'd go with english subdivision names, and consider having additional localized data if needed.

pmor:
Closing this in favour of #758, which uses ISO3166-2 data for subdivision types. I'll look into adding a method to "humanize" the types

PR #654, opened 2020-12-15
Fix missing Hebrew translation for CZ entry

hananamar:


hananamar:
The Hebrew translation for Czech Republic was missing.

rposborne:
Please make this change in the translation corrections file instead of anything under cache.

PR #650, opened 2020-11-13
Update i18n_data to 0.13.0 and drop support for rubies <2.5

pmor:

PR #656, opened 2021-01-03
Add Github actions on top of PR #650

petergoldstein:
Add Github actions on top of PR #650

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,10 @@
 cache: bundler
 language: ruby
 rvm:
- - 1.9.3
- - 2.3.6
- - 2.4.3
- - 2.5.0
- - ruby-head
+ - 2.5.8 
+ - 2.6.6
+ - 2.7,2
+ - 3.0.0 
 before_install:
   - gem update --system
   - gem install bundler
-matrix:
-  allow_failures:
-    - rvm: ruby-head
-    - rvm: 1.9.3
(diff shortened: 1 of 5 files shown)


petergoldstein:
@rposborne Travis.org is going away, so CI needs to transition to another system.  I threw together Github Actions, but for reasons that aren't clear they aren't running (essentially the same config is running on the logstasher repo)

.travis.yml
```diff
@@ -1,15 +1,10 @@
...
 language: ruby
 rvm:
- - 1.9.3
- - 2.3.6
- - 2.4.3
- - 2.5.0
- - ruby-head
+ - 2.5.8 
+ - 2.6.6
+ - 2.7,2
```
pmor, line 6: You have a comma here, that's probably what's breaking travis

pmor:
If I understand correctly, travis-ci.org will be replaced with travis-ci.com which continues to have a free plan for open source, but I can't find any details on what changes will be required or what limitations the new service might have, so I'm not sure we'll need to replace Travis.

Having said that, I have no objection with moving to GitHub actions

petergoldstein:
@pmor Thanks for pointing out the comma.

Re: Travis, my understanding is that the situation is a little more ambiguous that that.  Either way, I'm not sure why Github Actions aren't running with the addition of a YAML file under `.github/workflows`

petergoldstein:
@rposborne @pmor Do you have any idea when we'll see an new gem release for `countries`?

PR #681, opened 2021-04-23
Adds North Korea to list of unofficial names of KP

talha-akram:
It is pretty common to call `Democratic People's Republic of Korea` `North Korea` yet it is not present in the list of unofficial name and if we try to find country data for `North Korea` we get `nil` instead which leads to unexpected errors. Adding North Korea to the list of unofficial names improves accuracy of the matches and also fixes these issues.


touseefmurtaza93:
@talha-akram thanks for the PR.

pmor (approved):

PR #697, opened 2021-08-19
Update New Zealand's continent to Zealandia

CameronStanton:
The continent returned for New Zealand is out of date with the latest scientific consensus, see: https://en.wikipedia.org/wiki/Zealandia#Classification_as_a_continent

> In 2017, a team of eleven geologists from New Zealand, New Caledonia, and Australia concluded that Zealandia fulfills all the requirements to be considered a submerged continent, rather than a microcontinent or continental fragment.[6] This verdict was widely covered by news media.

### DIFF
--- a/lib/countries/data/countries/NZ.yaml
+++ b/lib/countries/data/countries/NZ.yaml
@@ -1,6 +1,6 @@
 ---
 NZ:
-  continent: Australia
+  continent: Zealandia
   address_format: |-
     {{recipient}}
     {{street}}


swiknaba:
That is cool, the main problem I see is, that this is a breaking change if I for instance rely on fetching all countries by Continent name, or if I've built business logic depending on the continent of a country since likely continent specific data/logic is hardcoded (because how often do continents change? :D)

pmor:
Closing this for now. As mentioned, it would be a potentially breaking change, and there doesn't seem to be a generalised consensus regarding this. 

None of the continental models of 4, 5, 6 and 7 continents that seem to be widely accepted include Zealandia, which is described in several sources as a micro-continent or submerged continent, which doesn't seem to be quite the same as a continent.

PR #666, opened 2021-02-25
Updating Norway subdivision codes to match late 2020 ISO change 

thomascrumrine:
As of 2020-11-24, the iso codes have been dramatically updated for Norway, see the bottom of [this page](https://www.iso.org/obp/ui/#iso:code:3166:NO) under "Change history of country code."

The translations were all pulled from https://query.wikidata.org

PR #704, opened 2021-11-29
Refresh subdivision data from CLDR

pmor:
run `update_cldr_subdivison_data` task to refresh subdivision data from CLDR


fschwahn:
Norway country codes were manually fixed in #666 - that change was overwritten with this update.

pmor:
Well spotted, I'll get it fixed soon.

PR #716, opened 2021-12-30
Docs: Changelog for 4.0.0: clarify breaking changes

jaredbeck:
Especially for major versions, it's helpful to those who are updating for us to explicitly state which changes are breaking, if any.


pmor:
Indeed, thanks for that!

PR #720, opened 2022-01-05
Remove non-ISO US subdivisions

pmor:
Remove US subdivisions AA, AE, AP. 

Armed forces subdivisions are USPS codes, but not ISO subdivisions

### DIFF
--- a/lib/countries/data/subdivisions/US.yaml
+++ b/lib/countries/data/subdivisions/US.yaml
@@ -1,32 +1,4 @@
 ---
-AA:
-  name: Armed Forces Americas
-  code:
-  unofficial_names: Armed Forces Americas
-  geo:
-    latitude: 33.8934985
-    longitude: -117.2732567
-    min_latitude: 33.8403275
-    min_longitude: -117.2985014
-    max_latitude: 33.9460679
-    max_longitude: -117.2498013
-  translations:
-    en: Armed Forces Americas
-  comments:
-AE:
-  name: Armed Forces Europe
-  code:
-  unofficial_names: Armed Forces Europe
-  geo:
-    latitude: 59.90441850000001
-    longitude: 10.7408094
-    min_latitude:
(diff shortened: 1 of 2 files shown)


henrik:
Hi! Stumbled over this PR/commit when some "AE" addresses in our system stopped working after a "countries" update.

Do you have any thoughts on what a suitable migration path is for pre-existing such addresses? A quick Google gives me https://kb.blackbaud.com/knowledgebase/Article/66988 which suggests picking California and adding "AE" to the city. I'll do a bit more research but thought I'd ask here as well.

pmor:
@henrik You can add custom data for subdivisions. It's not documented, but something like this in an initializer should let you add the non-ISO subdivisions:

```ruby
c = ISO3166::Country.new('US')
c.subdivisions['AA'] = ISO3166::Subdivision.new(name: 'Armed Forces Americas', code: 'AA')
c.subdivisions['AE'] = ISO3166::Subdivision.new(name: 'Armed Forces Europe', code: 'AE')
c.subdivisions['AP'] = ISO3166::Subdivision.new(name: 'Armed Forces Pacific', code: 'AP')
```

henrik:
Thank you! Yeah, that was one of the alternatives I considered. Perhaps that's a better idea in our case than migrating addresses to something ISO-valid. Not aware of us seeing any issues from non-ISO US subdivisions.

jaredbeck:
Don't forget to add translations,

```ruby
country.subdivisions['AE'] = ISO3166::Subdivision.new(
  name: 'Armed Forces Europe',
  code: 'AE',
  translations: {
    'en' => 'Armed Forces Europe'
  }
)
```

Methods like `subdivision_names_with_codes` will raise an error if translations are undefined.

pmor:
@jaredbeck Thanks for that, I hadn't noticed, and that sounds like a bug. Will look into it

PR #724, opened 2022-01-06
Missing alternative names for GB and US

dima4p:


pmor:
Thanks for the PR!

Could you please remove the changes to `Rakefile` and `lib/countries/cache/countries.json`? 
We regenerate the cache before a new release, otherwise it's a constant source of merge conflicts, and the Rakefile change appears to be unrelated to this PR.

dima4p:
> Could you please remove the changes to `Rakefile` and `lib/countries/cache/countries.json`? We regenerate the cache before a new release, otherwise it's a constant source of merge conflicts, and the Rakefile change appears to be unrelated to this PR.

With `lib/countries/cache/countries.json` it is clear, but what is wrong with  `Rakefile`? The change is to keep the order of the keys in `lib/countries/cache/countries.json`.

pmor:
Ok, the Rakefile change makes sense, but please see https://github.com/countries/countries#please-do-not-submit-pull-requests-on-cache regarding `lib/countries/cache/countries.json`

dima4p:
Additionally, reverting `lib/countries/cache/countries.json` breaks the specs…

PR #730, opened 2022-01-07
add VAT for Israel and Saudi Arabia

Venca24:
Adds 17 % VAT rate for Israel and 15 % for Saudi Arabia.

Sources:
- https://www.haaretz.com/israel-news/business/.premium-value-added-tax-is-down-as-of-today-but-its-not-out-1.5404351
- https://www.bbc.com/news/business-52612785


pmor:
Thanks for the PR!

PR #734, opened 2022-01-11
Update postal code regexes from rake task

cover:
A couple of postal code regexes got recently updated (fetched from `rake postal_codes:update`)

The diff for Israel shows that it's removed, but that's because it was duplicated (line 47-48 in the `IL.yaml` file)


pmor:
Well spotted, thanks!

PR #742, opened 2022-03-14
update i18n_data

bonekost:


afdev82:
Any news on this?
Could we relax the dependency on `i18n_data` to avoid to change every time the minor version?
Something like that maybe:

```
gem.add_dependency('i18n_data', '< 1')
```

pmor:
I'll try to release this later today. I'd rather not relax the dependency, because we've had bad data from the `salsa` repo get into `i18n_data` and then into our translations

afdev82:
OK, thanks.

I thought that probably there was a reason to not relax the dependency, thank you for the explanation.

PR #747, opened 2022-04-05
Added subdivision_names_with_codes method in documentation

arymoraes:
I added the `subdivision_with_names_with_codes` method to the documentation with a small example of the output since I thought it might be useful to know it exists.


pmor:
Thanks!

PR #748, opened 2022-04-18
Corrected Portuguese subdivision names

FilipaDurao:
Corrected the Portuguese translation of the Portuguese subdivisions.

The "Distrito de" before the district name was inconsistent (only appeared in some of the districts) and is unnecessary and redundant.


GuilhermeSousa95:
@pmor Would it be possible to merge this PR ? It correctly fixes inconsistencies regarding regions and districts of Portugal.

PR #763, opened 2022-07-15
Add an alternative name for USVI lookup

tadejm:
Looking up for an address in Virgin Islands US using Google Maps API responds with U.S. Virgin Islands as the country name:
    
```json
{
  "long_name" : "U.S. Virgin Islands",
  "short_name" : "VI",
  "types" : [ "country", "political" ]
},
```

CleanShot 2022-07-15 at 11 53 12@2x
 
This country name isn't specified as an unofficial name so lookup fails:
    
```ruby
[1] pry(main)> IS
O3166::Country.find_country_by_any_name("U.S. Virgin Islands")
=> nil
```


pmor:
Thanks for the PR!

PR #767, opened 2022-08-28
Replace sixarm_ruby_unaccent with unaccent

jonian:
This replaces [sixarm_ruby_unaccent](https://github.com/SixArm/sixarm_ruby_unaccent) with [unaccent](https://github.com/hardpixel/unaccent) (Disclaimer: I'm the author of unaccent). Unaccent uses the same accentmap as sixarm_ruby_unaccent so there will be not differences in the final output.

What it does differently is to auto-load the accentmap and use gsub to replace the accented characters. This results in faster conversions and less memory usage as you can see [here](https://github.com/hardpixel/unaccent#benchmark). Also the string extension is optional.

### DIFF
--- a/countries.gemspec
+++ b/countries.gemspec
@@ -23,7 +23,7 @@ Gem::Specification.new do |gem|
 
   gem.required_ruby_version = '>= 2.7'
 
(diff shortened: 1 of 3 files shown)


pmor:
Thanks for PR!

PR #779, opened 2022-11-12
Remove "Province of China" from Taiwan's short name

kevinluo201:
That's just not the fact at all : )

### DIFF
--- a/lib/countries/data/countries/TW.yaml
+++ b/lib/countries/data/countries/TW.yaml
@@ -28,7 +28,7 @@ TW:
   international_prefix: "002"
   ioc: TPE
   iso_long_name: The Republic of China
-  iso_short_name: Taiwan, Province of China
+  iso_short_name: Taiwan
   languages_official:
     - zh
   languages_spoken:


kevinluo201:
Please also check out [https://github.com/countries/countries/pull/29](https://github.com/countries/countries/pull/29)

pmor:
Thanks for the PR, but I'm going to have to reject it.

This gem has (for several years, as far as I know) stuck to a policy of being based on ISO3166 data in matters of country names and other issues, such as the independence of Kosovo.

Regardless of what the commonly used name of Taiwan is today, the fact is ISO3166 defines the official short name as  `Taiwan (Province of China)`  - https://www.iso.org/obp/ui/#iso:code:3166:TW

I suggest you use the `#common_name` method instead, since that will return the commonly used country names in English, instead of the ISO names. (eg, `Taiwan` instead of `Taiwan (Province of China)`, or `United Kingdom` instead of `United Kingdom of Great Britain and Northern Ireland`)

PR #781, opened 2022-11-30
Document how to get country codes (alpha2) [ci skip]

jean-francois-labbe:
It took me some time to figure out how to get an array of country codes. 
So I'd rather make it visible. 
My use case: I want to validate a country code user input against `ISO3166::Country.codes`.

Country codes are also called alpha2 in the documentation. This made it harder to find that the `ISO3166::Country.codes` method actually returns alpha2.


pmor:
Thanks for the PR!

PR #794, opened 2023-02-14
fix: RUP are part of the European Union

btrd:
Source https://fr.wikipedia.org/wiki/R%C3%A9gion_ultrap%C3%A9riph%C3%A9rique

It seems there is no English page about this on Wikipedia :/


btrd:
Also only some of them have a ISO3166 code. Don't hesitate to tell me if I can help in any way.

pmor:
FYI, I'm creating a new branch to integrate your PR with some additional work to better deal with these territories

PR #804, opened 2023-05-17
Add Japanese translation to subdivisions data

shucream-p:
Added Japanese translation to some subdivisions data files.


pmor:
Thanks for the PR! ❤️

PR #805, opened 2023-06-01
Return Taiwan and UK changes

800a7b32:
Clients require using non-standard (iso) names for both Taiwan and UK


800a7b32:
should not be merged to upstream

PR #820, opened 2023-09-06
Add distance_unit attribute to countries data

dwtong:
## Description
This adds the distance unit (MI/KM) for roading distance and speed for each country. Data for countries using miles (MI) is based on https://en.wikipedia.org/wiki/Mile#International. All other countries are assumed to use kilometres (KM). 

### Countries using miles (MI)
- Liberia
- Myanmar
- United Kingdom
- United States 

### Territories using miles (MI)
- American Samoa
- Bahamas
- Belize
- British Virgin Islands
- Cayman Islands
- Dominica
- Falkland Islands
- Grenada
- Guam
- N. Mariana Islands
- Samoa
- St. Lucia
- St. Vincent & The Grenadines
- St. Helena
- St. Kitts & Nevis
- Turks & Caicos Islands
- U.S. Virgin Islands


dwtong:
The spec that I introduced is failing as `rake update_cache` needs to be run. In the contribution guide it said not to include this in PRs? Also I am unsure why the other spec is failing?

PR #822, opened 2023-10-06
Remove duplicate key in subdivisions/US.yaml

pioz:
In the file `lib/countries/data/subdivisions/US.yaml` under "GA/translations" the key "tk" is duplicated. This can cause problems with libraries that do strict checks like [`go-yaml.v3`](https://github.com/go-yaml/yaml).
This PR removes the duplicate key.


pmor:
Thanks for the PR!

PR #828, opened 2023-11-30
add more national length to PS

shuaibzahda:
The phone number used in Palestine is the same as in Israel, so when using validating mobile numbers, the national number length shall include 9


pmor:
Thanks for the PR!

PR #831, opened 2024-01-06
fix: Corrected the Japanese name of Kyrgyzstan to the official name of the country

rjtt17:
Thank you for creating a great gem that is very easy to use!

I noticed that the Japanese spelling of the country name of Kyrgyz is the old name of the country. Therefore, I have created a pull request to correct the Japanese spelling of the country name of Kyrgyz to the official name. I refer to the following link for the Japanese spelling of the country name of Kyrgyz.

https://salsa.debian.org/iso-codes-team/iso-codes/-/blob/main/iso_3166-1/ja.po#L871
https://www.mofa.go.jp/mofaj/area/kyrgyz/data.html#section1

I hope you can check it when you have time.


pmor:
Thanks for the PR!

PR #833, opened 2024-01-21
Freeze Strings

swiknaba:
At work, we've been using this gem at scale (serving ~ 20-100M req/month), and detected this gem to pollute memory by allocating way too many strings.

We've been running this adjustment of freezing multiple strings for six months now without issues (and measurably lower memory footprint).

Sadly, I couldn't find our exact memory profiling results anymore (we did extensive before/after tests using the memory profiler gem and observing memory footprint on the servers).

I remember, that strings for country codes got allocated multiple times before, and only once after freeze strings.

Thank you for maintaining open source :-)


pmor:
Good one 👍, thanks for the PR

PR #834, opened 2024-01-21
remove unused gem retryable

swiknaba:
Minor cleanup; it seems this gem is not used anywhere.

Bundle install did a minor bump of nokogiri, that should be OK I guess.


pmor:
Well spotted, I missed that during a cleanup of old tasks we weren't using anymore. Thanks!

swiknaba:
Thanks for fixing the nokogiri version!

PR #835, opened 2024-02-07
Fix documentation for languages with dashes

sunny:
Minor documentation in the README to add an example of configuring languages with dashes `pt_BR`, `zh_CN`, `zh_TW`.


pmor:
Thanks for the PR!

sunny:
Thank you for maintaining the gem ❤️

PR #838, opened 2024-02-16
Update language and reduced VAT rate for Czechia

cervm:
This PR contains updates to languages and VAT rates for Czechia.

The only official language in Czechia is Czech. Slovak is also no longer considered a spoken language as it's only spoken by the Slovak minority nowadays.

https://european-union.europa.eu/principles-countries-history/country-profiles/czechia_en

As of 1 January 2024, the reduced VAT rate was changed to 12%.

https://www.globalvatcompliance.com/vat-rates-in-czech-republic/


pmor:
Thanks for the PR!

PR #847, opened 2024-05-14
Add conversion methods for country codes

mechnicov:
Useful methods to conversion from alpha-2 to alpha-3 and vice versa


pmor:
Quite useful, thanks for the PR.

Ignore those reek issues, I don't see anything wrong with the naming you picked.

mechnicov:
@pmor 
Naming issue is a very delicate matter even for humans. It's too much for machines

PR #849, opened 2024-05-17
Update EE VAT from 20 to 22 (01/01/2024 change)

jarthod:
Hello,

Very small PR, the Estonia (EE) standard VAT rate changed from 20% to 22% on January 1st, 2024.
Here is the official accouncement: https://www.emta.ee/en/business-client/taxes-and-payment/value-added-tax#from-01012024
I noticed this because I checked the European VAT table (which is up to date): https://europa.eu/youreurope/business/taxation/vat/vat-rules-rates/index_en.htm
I noticed it was not matching what I had in my code (which is based on countries).

Thanks!
Adrien

### DIFF
--- a/lib/countries/data/countries/EE.yaml
+++ b/lib/countries/data/countries/EE.yaml
@@ -51,7 +51,7 @@ EE:
   - Estonie
   - エストニア
   vat_rates:
-    standard: 20
+    standard: 22
     reduced:
     - 9
     super_reduced:


jarthod:
Just for the record, I also found another different about Spain rates (22% in EU website while 21% everywhere else), but for this one I couldn't find any confirmation and even the [official spain website](https://sede.agenciatributaria.gob.es/Sede/en_gb/iva/calculo-iva-repercutido-clientes/tipos-impositivos-iva.html) says it's 21%... So I believe this might be a mistake from them and I reported it to the EU website instead.

pmor:
Thanks for the PR!

Spain is definitely 21%, I just checked in the Spanish tax agency website.

jarthod:
Thanks for the confirmation, so there is definitely a mistake on the EU website for spain ^^
And thanks for the merge :)

PR #850, opened 2024-05-21
Change VAT rates for CH valid from 2024-01-01

phylor:
Switzerland changed their VAT rates by beginning of 2024.

Reference:

https://www.estv.admin.ch/estv/en/home/value-added-tax/vat-rates-switzerland.html

image


pmor:
Thanks for the PR!

PR #853, opened 2024-06-18
Update PY.yaml

jaredwindover:
Have been informed by customers, and cursory research verifies that Paraguay's postal code is 6 digits, not 4.


pmor:
Thanks for the PR!

PR #856, opened 2024-07-08
Rename countries with comma

younn-o:


pmor:
Thanks for the PR!

Could you please remove all changes made to the files in `/cache/`? Those files are generated before the gem is released.

younn-o:
Hey @pmor, the changes on the cache files were removed from the PR. Let me know if there is anything else to adapt, thanks !

pmor:
Apologies for the confusion. I did not mean to merge this PR.

I'm unsure if this is a good change or not, actually. Moving "Congo, Democratic Republic" to "Democratic Republic of Congo" will change the ordering of countries in a dropdown box, for instance, and make the distinguishing part of the name ("Congo") less visible, and the select box behave differently when using the keyboard to jump to the first letter of the country.

My very unscientific feeling is that most people will expect "Congo" and "Iran", instead of "Democratic Republic of Congo" or "Islamic Republic of Iran", and this change affects the "common name" output

PR #860, opened 2024-07-18
Correct Peterborough

kylewelsby:
Corrects the translation name for "Peterborough"

https://en.wikipedia.org/wiki/Peterborough


pmor:
Thanks for the PR!

PR #862, opened 2024-08-07
Adds new #all_names_with_alpha_3_codes method and spec

ifasoldt:
Hi! 

Really appreciate this gem. Had a need for a list of all countries with their alpha 3 codes and implemented in my app and then realized that it might make sense as a PR to here.

I implemented as a separate method from the existing `all_countries_with_codes` method, but I could also see it as an additional option to the existing method. 

```
 def all_names_with_codes(locale = 'en', code=alpha2)
   ... 
 end 
 
 ```

But I felt like a completely new method was probably the safest, even though theoretically adding an additional optional arg should not be risky. 

I also added a few additional expectations to the existing `all_countries_with_codes` specs to differentiate between its outputs and the new `all_countries_with_alpha_3_codes` outputs.

LMK if I need to make any changes.

### DIFF
--- a/lib/countries/country/class_methods.rb
+++ b/lib/countries/country/class_methods.rb
@@ -39,6 +39,13 @@ module ISO3166
       end.sort
     end
 
+    def all_names_with_alpha_3_codes(locale = 'en')
+      Country.all.map do |c|
+        lc = (c.translation(locale) || c.iso_short_name)
+        [lc.respond_to?(:html_safe) ? lc.html_safe : lc, c.alpha3]
+      end.sort
+    end
+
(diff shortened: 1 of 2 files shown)


ifasoldt:
I see that code climate has failed my PR. The errors are a result of me following the convention of the original `all_countries_with_codes` method for consistency. Would you like me to fix to make codeclimate happy or leave as is?

pmor:
Hi, thanks for the PR!

I'm currently travelling without a computer, so I can't really look at the code that well, but leave the code climate issues for me. I need to decide on the separate method vs your suggestion of an additional option.

PR #864, opened 2024-09-03
VAT rate for Finland changed to 25.5% as of 2024-09-01

JanMSP:
Finland changed their general VAT rate to 25.5 as of 2024-09-01.
Source: https://www.vero.fi/en/businesses-and-corporations/taxes-and-charges/vat/rates-of-vat/


pmor:
Thanks for the PR!

PR #865, opened 2024-09-06
Lazy load default locales

lacco:
While analyzing my project's load times, I realized that the country initializer took almost a second:

```
start = Time.now;

ISO3166.configure do |config|
  config.locales = %i[en fr]
end

puts "Time elapsed #{(Time.now - start) * 1000} ms"
```

I was able to break it down to this line:

https://github.com/countries/countries/blob/master/lib/countries/configuration.rb#L25

In my project, `I18n.available_locales` is being set later during the initialization.
Because of that, the default `I18n.available_locales` are being loaded, which take some time to initialize.

While in most projects `I18n.available_locales` should be set earlier,
just lazily loading the default locales might be a good improvement anyway.

```
without lazy default locales: Time elapsed 965.10 ms

with lazy default locales: Time elapsed 0.006 ms
```

### DIFF
--- a/spec/configuration_spec.rb
+++ b/spec/configuration_spec.rb
@@ -4,10 +4,6 @@ require 'spec_helper'
 require 'i18n'
 
 describe 'ISO3166.configuration' do
-  it 'has a configuration' do
-    expect(ISO3166.configuration).to be_a ISO3166::Configuration
-  end
-
   it 'locales can be changed' do
     ISO3166.configuration.locales = [:es]
     ISO3166.configuration.locales << :de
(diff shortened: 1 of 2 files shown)


spec/configuration_spec.rb
```diff
@@ -57,4 +57,18 @@
...
+
+      ISO3166::Configuration.new
+    end
+
+    it 'loads default locales when calling #locales for the first time' do
+      expect(I18n).to receive(:available_locales).once.and_return([:test])
+
+      expect(ISO3166::Configuration.new.locales).to eq([:test])
+    end
+  end
```
lacco, line 73: I don't know if those tests are needed.
They helped me to make the changes, but they might not provide long-term value.

Happy to remove them.

pmor:
Thanks for the PR!

Sounds like a great improvement, I should have some time next week to look at it.

pmor:
Last few weeks were really busy but, but this makes 100% sense. No point loading the default locales right away, as the configuration might set a completely different set of locales.

Issue #863, opened 2024-08-30
method_missing throws an internal error in ISO3166::CountryFinderMethods.method_missing


CoderMiguel:
```console
NoMethodError: undefined method `[]' for nil:NilClass

      return_all = matches[1]
                          ^^^
from /usr/local/bundle/gems/countries-5.6.0/lib/countries/country/finder_methods.rb:32:in `method_missing'
```
I have isolated the problem to this method:
https://github.com/countries/countries/blob/814b68757a90ef6ef4a2603a24ae57fb597dd291/lib/countries/country/finder_methods.rb#L30

I would recommend moving the guard clause up one line.

```ruby
    def method_missing(method_name, *arguments)
      matches = method_name.to_s.match(FIND_BY_REGEX)
      super unless matches
      return_all = matches[1] # <= this throws an error
      # super unless matches

      countries = find_by(matches[3], arguments[0], matches[2])
      return_all ? countries : countries.last
    end
```

This change works in a patch that I have made locally

PR #866, opened 2024-09-13
fix and tests

CoderMiguel:
Fixes https://github.com/countries/countries/issues/863

### DIFF
--- a/lib/countries/country/finder_methods.rb
+++ b/lib/countries/country/finder_methods.rb
@@ -29,8 +29,8 @@ module ISO3166
 
     def method_missing(method_name, *arguments)
       matches = method_name.to_s.match(FIND_BY_REGEX)
-      return_all = matches[1]
       super unless matches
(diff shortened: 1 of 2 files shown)


CoderMiguel:
The main branch of the fork is throwing an error locally:
Failures:

  1) ISO3166::Country subdivisions #states should be deprecated
     Failure/Error: expect { country.states }.to output.to_stderr
       expected block to output to stderr, but did not
     # ./spec/country_spec.rb:254:in `block (3 levels) in <top (required)>'
     
The reason for this failing is not obvious and does not seem to be coming from the changes I have introduced

pmor:
Thanks for the PR!

As for the test failure, if memory serves me, you'd need to run rspec with `-w` to get the deprecation warning specs to work

CoderMiguel:
No problem, this gem has been super helpful so far.  The least we can do is contribute back.

also, TIL about `-w`, Thx

PR #869, opened 2024-09-29
Remove deprecated `#state` method; Fix rubocop issues

pmor:


pmor:
3.0 test failures ignored as support will be dropped in the next release with #868

PR #871, opened 2024-10-23
Add vehicle_registration_code attribute to countries data

ch000:
This adds the vehicle registration code for each country which indicates in which country a motor vehicle's vehicle registration plate was issued. Data for countries is based on https://de.wikipedia.org/wiki/Liste_der_Kfz-Nationalit%C3%A4tszeichen#Liste_der_aktuellen_Kennzeichen.


ch000:
I think i'm missing something.

The tests ran successful and this works:
```
ISO3166::Country.new("de").vehicle_registration_code
=> "D"
```

But i'm getting an error here:
```
ISO3166::Country.pluck(:alpha2, :iso_short_name, :vehicle_registration_code)
KeyError: key not found: "vehicle_registration_code"
```

Is it maybe because i couldn't find vehicle registration codes for all lib/countries/data/contries/x.yaml files?

pmor:
Tests fail due to outdated cache file. Merging and will fix any issues that might still be present

PR #874, opened 2024-12-04
Update Federal Subdivisions of Nepal with New Provinces

himalayan-sanjeev:
This PR updates the federal subdivisions of Nepal to include the new provinces (Koshi, Madhesh, Bagmati, Gandaki, Lumbini, Karnali and Sudurpashchim) as per the new federal structure. It also retains the previous development regions and zones for historical purposes.

### Changes Made:
- Added 7 new provinces with their official names as `type: province`.
- Updated YAML structure with geo-coordinates, translations, and codes for provinces.

### DIFF
--- a/lib/countries/data/subdivisions/NP.yaml
+++ b/lib/countries/data/subdivisions/NP.yaml
@@ -797,78 +797,115 @@ NA:
   comments:
   type: zone
 P1:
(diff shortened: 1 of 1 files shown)


pmor:
Thanks for the PR!

PR #884, opened 2025-01-04
Experiment with symbol-keyed translation hashes to reduce string allocations

### LINKED ISSUE #872: Excessive allocations in translation
We have a few contexts (batch jobs) where we end up invoking [translation](https://github.com/countries/countries/blob/3639158bf98d2860c07f461e6af80d507a1e835f/lib/countries/country.rb#L106) a _lot_

I'm not sure that this is a _real_ problem, might just be user error but e.g. locally after a single batch job I can see:

```
Allocated String Report
-----------------------------------
    104534  "en"
     99750  /Users/tobymurray/.asdf/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/countries-7.0.0/lib/countries/country.rb:107
      3188  /Users/tobymurray/.asdf/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/countries-7.0.0/lib/countries/data/locales_methods.rb:16
      1594  /Users/tobymurray/.asdf/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/countries-7.0.0/lib/countries/data/locales_methods.rb:20
         2  /Users/tobymurray/.asdf/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/countries-7.0.0/lib/countries/data/subdivision_methods.rb:7
```
I'm sure we can update our code to figure out where this is getting invoked and avoid it, but it seems like `locale.to_s.downcase` is a bit of a subtle killer.

I've monkey patched it locally (we're only using English) with:
```
module ISO3166
  class Country
    def translation(locale = "en")
      locale = locale.to_s.downcase unless locale == "en"
      data["translations"][locale]
    end
  end
end
```

And confirmed it gets rid of the observed allocations, so it does feel like the culprit.

pmor:
Trying to fix #872 

This PR has potentially breaking changes.

Change the translation arrays in both countries and subdivisions to use Symbol keys instead of String keys and update code to minimize allocations when sanitizing and/or downcasing locale strings.

Access to translation arrays and calls to the translation methods should still work with both String and Symbol locales, but the actual arrays will use Symbol keys, and some methods will return these arrays, which can be a breaking change.

### DIFF
--- a/lib/countries/country.rb
+++ b/lib/countries/country.rb
@@ -108,8 +108,15 @@ module ISO3166
 
     # @param locale [String] The locale to use for translations.
     # @return [String] the name of this Country in the selected locale.
-    def translation(locale = 'en')
-      data['translations'][locale.to_s.downcase]
+    def translation(locale = :en)
+      case locale
+      when Symbol
+        locale = locale.downcase
+      when String
+        locale = locale.downcase if locale.match?(/[A-Z]/)
+        locale = locale.to_sym
+      end
+      data['translations'][locale]
     end
 
     # @return [String] the “common name” of this Country in English.
--- a/lib/countries/country/class_methods.rb
+++ b/lib/countries/country/class_methods.rb
@@ -18,7 +18,12 @@ module ISO3166
 
   module CountryClassMethods
     def new(country_data)
-      super if country_data.is_a?(Hash) || codes.include?(country_data.to_s.upcase)
+      unless country_data.is_a?(Hash)
+        country_data = country_data.to_s if country_data.is_a?(Symbol)
+        country_data = country_data.upcase if country_data.match?(/[a-z]/)
+      end
+
+      super if country_data.is_a?(Hash) || codes.include?(country_data)
     end
 
     def codes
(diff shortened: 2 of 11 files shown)


klyonrad:
@pmor This seems to completely change the return value of `#translations` method, outright returning something else than the documentation in the README.md file says

```ruby
 I18n.with_locale(:de) do
  country_obj = ISO3166::Country.find_country_by_alpha3("DEU")
  locale = :de
  country_obj.translations.fetch(locale.to_s)
end
```

the full hash of translations was

in 7.1 : `{"en"=>"Germany", "de"=>"Deutschland"}`

and in 8.0.1 : {:en=>"Germany", :de=>"Deutschland"}

----

Is that the idea? because then the README should get updated...

KieranP:
@pmor It would be good to add validations or a conversion for some methods. We just had issues because `country.subdivision_names_with_codes('en')` no longer works, and for India, was returning a mix of english and indian state names. Can this method (and others like it) convert strings to symbols automatically (though that could be a possible DOS attack point), or at least raise an error, stating that symbols are expected?

PR #890, opened 2025-02-21
fix(data): DK `address_format`

mroach:
The address format for Denmark was including `region`. This is not used in Danish addresses.

I ran into this whilst trying to validate address data in our database, ensuring we had all required components for rendering addresses.

Reference:

https://www.upu.int/UPU/media/upu/PostalEntitiesFiles/addressingUnit/dnkEn.pdf

https://www.postnord.dk/en/preparations/practical-information-about-letters-and-envelopes/addressing/

Additional reference: I live in Denmark and you never see any "region" data in an address. It's not relevant. It'd be like writing "Southern California" as a region for a US address.

### DIFF
--- a/lib/countries/data/countries/DK.yaml
+++ b/lib/countries/data/countries/DK.yaml
@@ -4,7 +4,6 @@ DK:
     {{recipient}}
     {{street}}
     {{postalcode}} {{city}}
-    {{region}}
     {{country}}
   alpha2: DK
   alpha3: DNK


mroach:
Where did the address formats come from in this gem? When building an address picker, it's important to know which countries require a subdivision (region). The closest I've come with this gem is:

```
ISO3166::Country.all
  .select { (_1.address_format || "").include?("region") }
  .map { [_1.alpha2, _1.iso_short_name] }
```

But that includes countries where it's not required or would never be used, like Singapore.

Are some of those address formats including `region` in the event that someone does want to include some region indicator?

PR #893, opened 2025-03-10
fix: the SK spec breakage

sabify:


pmor:
Thanks for the PR!

sabify:
Can we have a patch release on this one?

PR #899, opened 2025-04-02
fix: currency of Curacao and Saint Maarten

sabify:
The ANG currency was withdrawn as of 2025-03 in Curacao and Saint Maarten.


pmor:
Thanks for the PR!

PR #906, opened 2025-05-14
Correct ZA province name

yenshirak:
The official name of the province is `North West`, not `North-West`.

Source: https://www.gov.za/provinces


pmor:
Thanks for the PR!

PR #907, opened 2025-05-19
fix: iso short names

sabify:


pmor:
Thanks for the PR, but I'm not going to accept it.

The ISO short name is taken from the ISO pages (Example for AE: https://www.iso.org/obp/ui/#iso:code:3166:AE) and this gem has a long established policy of following ISO data, particularly for the `iso_long_name` and `iso_short_name` attributes. I'm not the original creator, but I do agree with that policy.

For most purposes, however, the names that are generally most useful are the ones you can find via the `translations(locale)` method, or, as a shortcut for the English locale, the `common_name` method (https://github.com/countries/countries?tab=readme-ov-file#names--translations)

sabify:
The corrections were based on the ISO 3166 short names, the same as your provided link. As the current data follows `ISO short name lower case`, this PR properly corrects the names.

If you are insistent on keeping the current state of the data, we can provide both fields that provide both `iso_short_name` (with upper case like how the ISO provides) and `iso_short_name_lower_case` (the other ISO provided field).

We can also see the ISO name with lower case in the wikipedia page: https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes

woodhull:
This is going to create a regression for me -- is there any way we could create a shortname field without the parenthesis / or articles?

We can strip this out in code, but having a clean field for public display purposes would be helpful.

PR #915, opened 2025-06-04
Add `iso_short_name_lower_case` attribute as an alternative to #907

pmor:


sabify:
@pmor I believe all countries should have the `iso_short_name` and `iso_short_name_lower_case` fields, and both fields should be copy-pasted of their ISO representation.

Please consider the UPPERCASE form of a country ISO short name may be different from the lowercase form of it. For example, [Turkey](https://www.iso.org/obp/ui/#iso:code:3166:TR) ISO short names are different in both forms. (It's different in the `i` character in both forms that is not convertible in many PLs).

I want you to please use `iso_short_name` and `iso_short_name_lower_case` exactly the same as it is indicated in the ISO website; these are identifiers. `iso_short_name` should be the UPPERCASE form, and the `iso_short_name_lower_case` should be the lowercase form.

Some may experience regression in their data, but we shall have the correct data.

sabify:
Please consider that based on the ISO data reference ([sample](https://www.iso.org/obp/ui/APP/global/0/legacy/0/iso_country_code_CH_csv.zip)), we need to keep the `iso_short_name` field and data as it is and just add another field for `iso_short_name_uppercase` for all countries.

PR #917, opened 2025-06-05
fix: add guard clause to check if matches is nil

ratheeshaditya:
Hi,

After the release of Country version 8, certain parts of our application began breaking with the following error:

```
   NoMethodError:
     undefined method `[]' for nil
   # ./lib/countries/country/finder_methods.rb:49:in `respond_to_missing?'
```

We traced the issue to the `matches` variable being `nil` and still being accessed with an index in `finder_methods`. To reproduce this, I’ve added a spec that demonstrates the failure.

This PR fixes the issue by adding a guard clause to return early, if `matches` is `nil` preventing the `NoMethodError`.

### Our Usage
<hr>

The behavior surfaced when using the gem in combination with the CanCanCan authorization gem. Internally, CanCanCan calls `respond_to?(:arr)` on the `ISO3166::Country` object. That call triggered the error shown above.

If this fix isn’t appropriate or there's a preferred way of doing it, kindly comment :)

### DIFF
--- a/lib/countries/country/finder_methods.rb
+++ b/lib/countries/country/finder_methods.rb
@@ -46,12 +46,12 @@ module ISO3166
     # :reek:BooleanParameter
     def respond_to_missing?(method_name, include_private = false)
       matches = method_name.to_s.match(FIND_BY_REGEX)
+
+      return super unless matches
+
       method = matches[3]
-      if matches && method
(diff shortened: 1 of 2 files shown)


pmor:
Thanks for the PR!

ratheeshaditya:
EDIT: I've added an additional condition to ensure the method returns a boolean

PR #919, opened 2025-06-12
fix: change french name of France subdivisions

Bertrand-Bichat:
In file `lib/countries/data/subdivisions/FR.yaml` : 
- use French name `Nouvelle-Calédonie` for `New Caledonia` as subdivision of France country

In file `lib/countries/data/translations/countries-fr.yaml` : 
- add `CP: Clipperton` because there is also `TF: Terres australes françaises` in the same file

### DIFF
--- a/lib/countries/data/subdivisions/FR.yaml
+++ b/lib/countries/data/subdivisions/FR.yaml
@@ -8430,7 +8430,7 @@ NAQ:
     sl: Nova Akvitanija
   type: metropolitan_region
 NC:
-  name: New Caledonia
+  name: Nouvelle-Calédonie
   code: NC
   unofficial_names:
   - Nouvelle-Calédonie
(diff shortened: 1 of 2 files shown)


pmor:
Hi @Bertrand-Bichat ,

Thanks for the PR!

We follow ISO naming as much as possible, and the name changes you have here do not match ISO data from https://www.iso.org/obp/ui/#iso:code:3166:FR , so I'll gladly accept the new translations, but I'll ask you to revert the changes to the `name` attributes, so that we keep the ISO data there.

Bertrand-Bichat:
> Hi @Bertrand-Bichat ,
> 
> Thanks for the PR!
> 
> We follow ISO naming as much as possible, and the name changes you have here do not match ISO data from https://www.iso.org/obp/ui/#iso:code:3166:FR , so I'll gladly accept the new translations, but I'll ask you to revert the changes to the `name` attributes, so that we keep the ISO data there.

@pmor ok, I have modified my commit to only change the New Caledonia French subdivision name attribute.
Is that correct for you ?

Bertrand-Bichat:
@pmor can you tell me if my pull request is correct for you, please ?

Issue #901, opened 2025-04-07
"zh" not supported?


jerry81:
Relevant parts of our code:

Gemfile.lock

    countries (5.6.0)
      unaccent (~> 0.3)

application.rb

    config.i18n.available_locales = %i[en de zh-CN zh]

puts ISO3166::Country.all.first.translations
=> {"en"=>"Andorra", "de"=>"Andorra", "zh"=>nil, "zh-cn"=>"安道尔"}

Is it possible to add zh support, since i see vanilla :en and :de don't have issues

pmor:
Yes, I suppose I can do that, but I don't know which would be the correct content to use for it. We currently have `zh_CN`, `zh_HK` and `zh_TW` locales.

jerry81:
i googled it and got

Simplified Chinese:
Also known as "zh-CN" (or sometimes just "zh"), is the script used in mainland China and Singapore. 
Traditional Chinese:
Also known as "zh-TW" or "zh-Hant", is used in Taiwan, Hong Kong, and Macau. 

but up to you, thanks for considering it

gordoncl:
> Yes, I suppose I can do that, but I don't know which would be the correct content to use for it. We currently have zh_CN, zh_HK and zh_TW locales.

This might merit opening a new issue (I'm going to piggy back off of this) but why not use the CLDR to determine the content?  From looking at the code it seems like this library already uses CLDR for getting translations for subdivisions. We could write a rake task which fetches the country translations, rather than rely on user pull requests.

**Aside**: Where I work, our vietnamese translator mentioned that the translations for territories in vietnamese are kind of off. And when I did a comparison I noticed the translations differed from the CLDR, which kind of surprised me.

The CLDR provides territory translations for `zh`, `zh-Hant` and `zh-Hans`:

* [zh](https://github.com/unicode-org/cldr/blob/4746f135c556756b79906cf3d3205d40e23ff92e/common/main/zh.xml#L825-L1137)
* [zh-Hant](https://github.com/unicode-org/cldr/blob/4746f135c556756b79906cf3d3205d40e23ff92e/common/main/zh_Hant.xml#L885-L1197)
* [zh-Hans](https://github.com/unicode-org/cldr/blob/4746f135c556756b79906cf3d3205d40e23ff92e/common/main/zh_Hans.xml)

[...]

PR #924, opened 2025-07-21
Subdivision tasks

gordoncl:
## Overview

In response to https://github.com/countries/countries/issues/901#issuecomment-3076673274, I decided to experiment a little bit with auto importing CLDR translations. Noticing we already had some scripts for importing translations for subdivisions I started to repurpose some of that code and found some bottlenecks.

In particular, I reworked the [download_folder](https://github.com/countries/countries/blob/a0674cea576727fdfd2e7f6430283950873540fc/lib/countries/sources/cldr/downloader.rb#L21-L34) method. The main reason I did this was because if we want to fetch all translations for territories, we need to be able to get files from the `main` directory. But the directory API only shows up to 1000 entries, which the `main` directory exceeds.

The new approach fetches the entire CLDR repository by downloading the latest zip file and extracting it's contents. This has a side benefit of also making the rake tasks much faster to run.

The last commits add the results of running the rake tasks to update the translations and also found that there are a lot of subdivisions we're missing information for. Updating ISO data feels out of scope so I didn't pursue adding that, but did update the [cleanup_subdivision_yaml](https://github.com/countries/countries/blob/a0674cea576727fdfd2e7f6430283950873540fc/Rakefile#L51-L62) task to remove bad entries such that the tests would pass.

### Chakma Script

While running the rake tasks I found that the `psyche` or `yaml` library encodes the `ccp` ([Chakma Script](https://en.wikipedia.org/wiki/Chakma_script)) as unicode escape sequences. I put these changes into a separate commit but I can drop it if we want. We can in theory include a script which post processes the files to encode as UTF-8, but I would honestly just delegate to what these libraries do. It's not clear to me if this encoding is environmental or at the library level. But something to note is Chakma Script lives on Supplementary Multilingual Plane, which might explain this kind of encoding.

### Unused locales

It seems like the country translations we have do not include all the locales the CLDR has. But, the subdivision translations we use are all encompassing. It's unclear to me whether we are choosing to support a limited number of locales or are trying to support as many locales as possible. Regardless, I just tried to preserve the essence of what the script was doing before.


lib/countries/data/subdivisions/AR.yaml
```diff
@@ -1802,4 +1813,10 @@ Z:
     yue: 聖古絲省
     mk: Санта Круз
     'no': Santa Cruz
+    az: Los Buzlaqları Milli Parkı
+    hr: Nacionalni park Los Glaciares
+    ml: ലോസ് ഗ്ലേസിയേഴ്സ് നാഷണൽ പാർക്ക്
+    my: လော့စ်ဂလက်ဆရားရက်စ် အမျိုးသားဥယျာဉ်
+    ne: लोस ग्लासिआरेस राष्ट्रिय निकुञ्ज
+    sq: Parku Kombëtar Los Glaciares
```
gordoncl, line 1826: This is an example of a locale which we don't have translations for in the country translations directory. As mentioned in the main comment I'm not sure whether we want a supported locale configuration or if we eventually want to include all translations.

lib/countries/data/subdivisions/AE.yaml
```diff
@@ -68,7 +68,7 @@ AJ:
     sr_Latn: Adžman
     sw: Ajmān
     yue_Hans: 阿治曼酋长国
-    ccp: "𑄃𑄌𑄴𑄟𑄚𑄴"
+    ccp: "\U00011103\U0001110C\U00011134\U0001111F\U0001111A\U00011134"
```
gordoncl, line 71: I mentioned this in the main comment but these are being encoded as escape sequenced unicode characters. These characters are on the [SMP](https://en.wikipedia.org/wiki/Plane_(Unicode)#Supplementary_Multilingual_Plane). This encoding occurs from calling `to_yaml`.

```ts
$ bundle exec irb
irb(main):001> require 'yaml'
=> true
irb(main):002> yaml = { field: "𑄃𑄌𑄴𑄟𑄚𑄴" }
=> {:field=>"𑄃𑄌𑄴𑄟𑄚𑄴"}
irb(main):003> yaml.to_yaml
=> "---\n:field: \"\\U00011103\\U0001110C\\U00011134\\U0001111F\\U0001111A\\U00011134\"\n"
irb(main):004> YAML.load(yaml.to_yaml)
=> {:field=>"𑄃𑄌𑄴𑄟𑄚𑄴"}
```

This likely means the encoding is coming from libyaml or psyche. We could add a post processing script (cleanup task) if wanted to encode this as unicode. But the information is properly decoded 🤷 

Would be interested to see if running the script on other environments produces similar output. It would be nice if the results were canonical. But that might be contingent on locking versions of `libyaml`.

pmor:
Hi @gordoncl,

Thanks for the PR. I will need some time to review this one, since I want to be reasonably sure the updated data is correct, and there's a lot of files changed.

I've stopped using the CLDR importer some time ago after it imported data that conflicted with ISO data for subdivisions. This gem has a policy of following ISO data for much longer than I've been its maintainer, and I'm not inclined to change it, so ISO data takes precedence over any other source.

lib/countries/data/subdivisions/AR.yaml
pmor: Which locale? I think we have these 6.

lib/countries/data/subdivisions/AE.yaml
pmor: The XML files from CLDR are unicode, so if we could preserve the output as unicode, I think it would be nicer. Also, it would eliminate a lot of changes from this PR that are caused by the encoding change.

Rakefile
```diff
@@ -48,15 +48,30 @@ task :update_cache do
...
     data = sd.load
     next if data.nil? || data == {}
 
     data = data.sort.to_h
     data['translations'] = data['translations'].sort.to_h unless data['translations'].nil?
-    sd.save(data)
+
+    # Remove incomplete entries
+    cleaned_data = data.select do |_, props|
+      props['name'] && props['code'] && props['type']
```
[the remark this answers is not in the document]
gordoncl: Thanks for doing some auditing.

Looking into this a bit it does seem like the CLDR keeps deprecated identifiers around. [Updating subdivision codes](https://cldr.unicode.org/development/updating-codes/updating-subdivision-codes) mentions never removing an identifier, with a link to a discussion about data stability.

> For instance, AL has 12 subdivisions total, so I don't understand where the 36 come from, unless it's severely outdated.

This shows 36 AL codes (in the english locale) and they have a comment indicating that these are deprecated.

https://github.com/unicode-org/cldr/blob/4746f135c556756b79906cf3d3205d40e23ff92e/common/subdivisions/en.xml#L95-L130

> PL has 16 subdivisions, but they changed the all codes in 2018, which matches the 16 incomplete entries you're seeing there

The deprecated PL subdivision translations seem to come from the infamous Chakma script translation file:

https://github.com/unicode-org/cldr/blob/4746f135c556756b79906cf3d3205d40e23ff92e/common/subdivisions/ccp.xml#L5422-L5438

There does appear to be good news, it seems like the CLDR provides meta data for detecting what is deprecated and what isn't in this file:

https://github.com/unicode-org/cldr/blob/4746f135c556756b79906cf3d3205d40e23ff92e/common/validity/subdivision.xml

What might make sense is when running the update rake task, to actually start by comparing our subdivision codes to the codes in the CLDR and prompt the user about expected discrepancies before updating the files. We can mention both what codes appear exclusively on our end and what codes appear exclusively on their end. 

Ultimately though, I think since the policy of the library is to treat ISO as the source of truth, no data should be pulled in unless it can somehow be mapped to the subdivision codes that we have. I initially accomplished this with a post processing script to remove extraneous data, but perhaps this should be done by just iterating through our subdivision codes in each translation file 🤔 

I'll look into making a fixup commit which makes these changes.

gordoncl (commented):
Hey @pmor. 

I made some revisions to the PR based on feedback. And I've also tried to split up some changes so it should be easier to audit what has changed (albeit there are still a lot of new translations for certain locales). 

There are three commits which update the yaml files.

* https://github.com/countries/countries/pull/924/commits/db35e6e9d09768f1067f2eecfffb97e39fe5f6c9 runs the code simply through the yaml parser to incorporate recent changes from `psych`
* https://github.com/countries/countries/pull/924/commits/db35e6e9d09768f1067f2eecfffb97e39fe5f6c9 just runs the `cleanup_subdivision_yaml` rake task, which sorts the keys in some of the files
* https://github.com/countries/countries/pull/924/commits/db35e6e9d09768f1067f2eecfffb97e39fe5f6c9 runs both the `update_cldr_subdivison_data` and `cleanup_subdivision_yaml` tasks

The last commit should only contain the actual changes in terms of translations. If you want, I can perhaps look into splitting that last commit into two other commits: one which highlights which translations were changed and which were added.

gordoncl:
I'm going to close this PR since a lot of time has passed since it's been opened and there hasn't been a lot of movement. I probably should have spent more time hashing out whether or not this was desirable in an issue before opening a PR, that is my mistake. 

From my perspective, I think it would be nice if the library offered a way to look up CLDR translations, even if the primary English translations reflect the ISO standard. The CLDR translations are designed to be used in software. And the ISO standard doesn't really provide any guidance outside of English (and French) as far as I know.

Since the library offers a way to override the provided translations, this is what my team has done.

PR #947, opened 2026-02-18
[~] avoid redundant objects allocations

skcc321:
Optimize finder methods to reduce object allocations. Especially helpful on "hot" code.

   ### Summary

   Reduces memory allocations in find_country_by_any_name and related finder
   methods by 98.6% (from 1,530 to 21 objects per call).

   ### Key Changes

     - Country instance caching - Cache and reuse Country objects by alpha2 code
     - Direct data hash access - Access value[attr] directly instead of calling methods that allocate new arrays
     - Cache searchable attributes - Memoize the searchable attributes array
     - Optimize array handling - Use direct array literals and frozen constants
     - Skip nil values - Prevent nil matching in array comparisons

   ### Testing

     - All 269 tests pass ✓
     - No breaking changes

   ### Verification
```ruby
$LOAD_PATH.unshift(File.expand_path('lib'))
require 'memory_profiler'
require 'countries'

ISO3166::Country.find_country_by_any_name('United States')

report = MemoryProfiler.report do
  100.times { ISO3166::Country.find_country_by_any_name('United States') }
end

puts "Allocated: #{report.total_allocated} objects"
puts "Avg: #{report.total_allocated / 100} objects per call"
```
   Expected output: ~21 objects per call (vs ~1,530 before optimization)

### DIFF
--- a/lib/countries/country/class_methods.rb
+++ b/lib/countries/country/class_methods.rb
@@ -26,6 +26,16 @@ module ISO3166
       super if codes.include?(country_code)
     end
 
+    def cached_instance(country_data)
+      @_country_instance_cache ||= {}
+      alpha2 = country_data['alpha2']
+      @_country_instance_cache[alpha2] ||= new(country_data)
+    end
+
+    def reset_country_cache
+      @_country_instance_cache = {}
+    end
+
     # :reek:UtilityFunction
     def codes
       ISO3166::Data.codes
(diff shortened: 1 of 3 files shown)


dkh (approved):

skcc321:
@pmor, hi there! Could you take a look at this PR, please?

pmor:
Thanks for the PR!

